PR #65, opened 2022-07-11
Create static code analsiys and dependabot workflows

shenanigansd:
Add CodeQL for static code analysis and create a dependabot config file to enable dependabot.

Please note that dependabot does not yet support PEP 621, and won't be able to see the Python dependencies until it gets updated. This is being tracked in https://github.com/dependabot/dependabot-core/issues/3290

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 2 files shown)


FeodorFitsner:
Hey @shenanigansd! Thanks for submitting this. Is it still actual PR? Before merging this is there any example to see how it works and how results look like? Is it possible to run locally?

shenanigansd:
Oh hey! Sorry, I meant to add more explanation to this one, but then I completely forgot about it.
Actually, they're both config files to enable GitHub services, so they can't be run locally.
The CodeQL analysis sets up a GitHub Actions workflow to scan each commit for security  issues; and Dependabot is a GitHub service that automatically creates PRs when your dependencies are out of date.
Here's an example, https://github.com/darbiadev/darbiadev-shipping/pull/7
Dependabot create the PR, and you can see the output of the CodeQL analysis in the Actions tab.

PR #139, opened 2022-08-03
Fix Container on_click callback

mikaelho:
Setter and getter event names did not match, causing `on_click` callbacks not to work.


FeodorFitsner:
Oh, good catch!

Issue #230, opened 2022-08-25
app doesn't have a CONTRIBUTING.md file.


techemmy:
Is this app open-source?
I love this project and I'd love to be a contributor but I can't find the CONTRIBUTING.md file

FeodorFitsner:
Thanks for your interest!

There is an open issue for creating contributing guide: https://github.com/flet-dev/flet/issues/165

What would you like to contribute?

techemmy:
I'm looking to contribute to the python part of the code since that's the only one I'm knowledgeable in out of all languages you use.

techemmy:
Do you mind if I move the CONTRIBUTING.md file to the main directory so that it'll be easy for new visitors to find it?

or perhaps, I should add a message and link of the CONTRIBUTING.md to the README.md file indicating how to contribute?

FeodorFitsner:
Sure, let's move it. Please do a PR.

once it's moved to the root I will add about Flutter part.

techemmy:
Okay

PR #231, opened 2022-08-26
I moved the CONTRIBUTING.md file to the project's main directory.

techemmy:
I also referenced the CONTRIBUTING.md file in the README.md


techemmy:
Resolves #230

FeodorFitsner:
As you moved `CONTRIBUTING.md` to a root of the repo I think we should remove it from `sdk/python`. Could you do that please? I suggest having sub-sections in the main `CONTRIBUTING.md` for different components/languages, such as "Flutter", "Python client", etc.

techemmy:
> As you moved `CONTRIBUTING.md` to a root of the repo I think we should remove it from `sdk/python`. Could you do that please? I suggest having sub-sections in the main `CONTRIBUTING.md` for different components/languages, such as "Flutter", "Python client", etc.

Sure, will do that now.

PR #236, opened 2022-08-27
Vector utility class

mikaelho:
This PR suggests that we add a simple immutable 2D vector class to the flet utils.

_Why_

When playing with fancy UIs and absolute positioning, I often find myself placing things in circular formations, or "this far from that in this direction". This is easy to think about and implement if you have a 2D vector class that allows you to adjust the angle or length of the vector independently, or just add vectors together (even though that is simple tuple addition).

For unknown reasons, Python stdlib does not include a vector class, and the internet and PyPI are full of implementations. Here is one more. Suggestion is to include it in the utils so that tutorials and samples do not need to add requirements for it.

_Usage_

Create and access - coordinates

```python
    >>> v = Vector(1, 2)
    >>> v.x, v.y
    (1.0, 2.0)
```

Create and access - angle and magnitude (length)

```python
    >>> v = Vector.polar(math.pi, 2)
    >>> v
    Vector(-2.0, 0.0)
    >>> v.magnitude  # Length of the vector, alias for abs(v)
    2.0
    >>> v.radians
    3.141592653589793
    >>> v.degrees
    180.0
```

Arithmetic operations

```python
    >>> Vector(1, 1) + 2
    Vector(3.0, 1.0)
    >>> Vector(0.1, 0.1) + Vector(0.2, 0.2) == Vector(0.3, 0.3)  # Float tolerance 10 decimals
    True
    >>> Vector(2, 3) - Vector(1, 1)
    Vector(1.0, 2.0)
    >>> Vector(1, 1) * 2
    Vector(2.0, 2.0)
    >>> round(Vector.polar(math.pi / 4, 1), 1)
    Vector(0.7, 0.7)
```

Get a new vector by adjusting angle or magnitude

```python
    >>> v = Vector(1, 2)
    >>> v = v.with_magnitude(4.47213595499958)  # Twice as long
    >>> v.x, v.y
    (2.0, 4.0)
    >>> v = Vector.polar(math.pi, 2)
    >>> v
    Vector(-2.0, 0.0)
    >>> v.with_radians(0)
    Vector(2.0, 0.0)
    >>> v.with_degrees(90)
    Vector(0.0, 2.0)
```

### DIFF
--- a/sdk/python/flet/utils.py
+++ b/sdk/python/flet/utils.py
@@ -1,3 +1,4 @@
+import math
 import os
 import platform
 import sys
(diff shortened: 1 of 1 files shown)


FeodorFitsner:
Looks good.

PR #251, opened 2022-08-30
Fixed on_close event handling

FI-Mihej:
There is an issue with "on close" event which leads to wrong behavior and to a memory leak.
`on_event` receives 'disconnect' instead of 'close' on both browser tab close and an app window close.

Reproduces on version: 0.1.54 from PyPi
OS: Windows 10

### DIFF
--- a/sdk/python/flet/flet.py
+++ b/sdk/python/flet/flet.py
@@ -186,7 +186,7 @@ def _connect_internal(
             conn.sessions[e.sessionID].on_event(
                 Event(e.eventTarget, e.eventName, e.eventData)
             )
-            if e.eventTarget == "page" and e.eventName == "close":
+            if e.eventTarget == "page" and (e.eventName in {"close", "disconnect"}):
                 print("Session closed:", e.sessionID)
                 del conn.sessions[e.sessionID]
 


FeodorFitsner:
Indeed, as session ID is tab-specific the session is essentially ended when a browser tab/window is closed.
But `disconnect` will be also called in case of network hickup when WebSocket client is disconnected. We definitely don't want to kill a session on re-connect. On the other hand Flet server will send `close` event when a session is expired (60 minutes by default but can be changed in config). So, eventually session's page object will be freed on Python side.

FI-Mihej:
@FeodorFitsner Hello! I saw not really happy comments on Reddit about the way page is constructed in Flet. So I've changed it in my wrapper (https://github.com/FI-Mihej/flet_async) half a year ago. Unfortunately it is not possible for me to spend time for a pull request creation or for syncing my code with a new Flet version currently (since I've just arrived to Canada and a quite busy as an average newcommer by preparing things around for my wife and our kids who still waiting for their CUAET visas). So I believe it would be great if you will look to my approach (or to an approach used in for example NiceGUI) and try to use it in Flet. 
Cheers :)

FeodorFitsner:
@FI-Mihej Slava Ukraine and welcome to Canada! :)

I believe the issue with memory leak in `on_close` was fixed in https://github.com/flet-dev/flet/issues/969 - take a look. There is also a way to control the lifetime of a user session with environment variable. When browser tab is closed or re-connection occurs (say, on computer unlock) a user session is not killed.

Async support is also built into Flet now: https://flet.dev/docs/guides/python/async-apps

FI-Mihej:
@FeodorFitsner Heroyam Slava! Thank you! :)

Yes. Actually I'm not talking about the topic of the Pull Request 251. I'm just reusing current conversation in order to make a few fast direct messages to you about the different topic. I'm sorry if it is inconvenient way :) 

I'm talking about the approach when the child item should be at least defined with Control Refs before parent will be able to be constructed. And as for example with Lisp-like languages, current approach can result to a Parenthesis Hell. So resulting code of real complex forms can be somehow more difficult to maintain this way. For example:

```python
import flet as ft

def main(page):
    first_name = ft.Ref[ft.TextField]()
    last_name = ft.Ref[ft.TextField]()
    greetings = ft.Ref[ft.Column]()

    def btn_click(e):
        greetings.current.controls.append(
            ft.Text(f"Hello, {first_name.current.value} {last_name.current.value}!")
        )
        ...

    page.add(
        ft.TextField(ref=first_name, label="First name", autofocus=True),
        ft.TextField(ref=last_name, label="Last name"),
        ft.ElevatedButton("Say hello!", on_click=btn_click),
        ft.Column(ref=greetings),
    )

ft.app(target=main)
```

1) An example my approach (part of the Page construction method):
```python
...
with root()(Row, alignment="center") as r:
    with r()(Column, alignment='center') as c:
        with c()(Container, width=400, alignment=flet.alignment.center) as con:
            self.input = con()(TextField, label="Input", hint_text="Please enter text here", on_change=self.input_on_change, multiline=True).item
            self.input2 = con()(TextField, label="Input 2", hint_text="Please enter text here", on_change=self.ainput_on_change, multiline=True).item
...
```

As you can see, the above code holds the same hierarchical structure the developer expects to see on the rendered page. Source code has no tendency to form a Parenthesis Hell. And in above code I'm using `self.input` and `self.input2` for holding text field. It is a bit more common method than derefferencings like `last_name.current`

2) An example of the NiceGUI's approach:
```python
from nicegui import ui

ui.icon('thumb_up')
ui.markdown('This is **Markdown**.')
ui.html('This is <strong>HTML</strong>.')
with ui.row():
    ui.label('CSS').style('color: #888; font-weight: bold')
    ui.label('Tailwind').classes('font-serif')
    ui.label('Quasar').classes('q-ml-xl')
ui.link('NiceGUI on GitHub', 'https://github.com/zauberzeug/nicegui')

ui.run()
```

I think that after a while I will have the opportunity to discuss this topic in more depth (because I will need to return my focus to it before talking). If this topic interests you, of course.

FeodorFitsner:
@FI-Mihej sure, will gladly discuss that pattern!

PR #264, opened 2022-09-02
Querystring extract

StanMathers:
Well, when doing some project with flet, I found I needed to get query string data as key-value pairs to send it to the backend and since there was no way to extract data from the query string, I wrote the code myself, so I think it's helpful and must be merged with flet. Moreover, it's needed when dealing with web.

QueryString class consists of 2 methods, one of them is called `get()` which takes `key` as an argument and returns its value accordingly, and another is `to_dict()` which converts querystring into a python dictionary. 

Also, there is docstring description in the class.

### DIFF
--- a/sdk/python/.gitignore
+++ b/sdk/python/.gitignore
@@ -136,3 +136,6 @@ __pypackages__/
 flet/bin/
 
 /playground
+
+# Testing area
+tests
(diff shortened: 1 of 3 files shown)


FeodorFitsner:
Thanks for your PR!

What about URL-encoded values in a query? I think it would be helpful if this helper URL-decode values when you get them?

StanMathers:
> 

Thanks for the feedback. That seems really good advice, I'll implement this feature soon.

adampiskorski:
I may be missing something here, but rather than using Regex, perhaps use urllib.parse?
https://stackoverflow.com/a/5075477/6251468

FeodorFitsner:
Right, there is a helper for that: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs

PR #272, opened 2022-09-05
add error fix for those working with a source package

techemmy:
Added how to fix an error when working with a source package


techemmy:
update README.md

PR #280, opened 2022-09-06
Added querystring parser

StanMathers:
Well, I accidentally deleted the previous pr, so this is mostly a finished version of query string parser.
Soon, I'll add examples to flet examples and get docs ready.

Currently, supported features are:
- page.query.get(key) -> gets value according to querystring key (../?q=5 -> page.query.get('q') -> 5)
- page.query.to_dict -> converts querystring into python dictionary
- page.query.path -> gets URL path

Changes with `go` function:
- Added **kwargs which is encoded into querystring.
    page.go('/something/', key1=val1, key2=val2...)

page.query.post() -> this is for internal use with go function and has nothing to do externally


FeodorFitsner:
FYI: CI is failing: https://ci.appveyor.com/project/flet-dev/flet/builds/44694994/job/d9o2avbg4d27u9gd

StanMathers:
> FYI: CI is failing: https://ci.appveyor.com/project/flet-dev/flet/builds/44694994/job/d9o2avbg4d27u9gd

Now it succeded, I had a really silly mistake I could never imagine I would do.

FeodorFitsner:
That what CI is for :)

sdk/python/flet/page.py
```diff
@@ -102,6 +103,9 @@ def convert_keyboard_event(e):
         self.__on_disconnect = EventHandler()
         self._add_event_handler("disconnect", self.__on_disconnect.handler)
 
+        # Querystring
+        self.query = QueryString(page=self)
```
FeodorFitsner, line 107: 1) looks like route is getting parsed on initial page load only? It needs to be parsed on `route_change` event too, right?
2) Let's use a private `page.__query` to store parsed query and then expose `page.query` as a property, like anything else.

sdk/python/flet/page.py
```diff
@@ -282,8 +286,10 @@ def show_signin(self, auth_providers="*", auth_groups=False, allow_dismiss=False
             elif e.control == self and e.name.lower() == "dismisssignin":
                 return False
 
-    def go(self, route):
-        self.route = route
+    def go(self, route, **kwargs):
+        self.querystring_data = self.query.post(kwargs)
+        self.route = route if kwargs == {} else route + self.querystring_data
```
FeodorFitsner, line 291: Could be just?

```python
        self.route = route if kwargs == {} else route + self.query.post(kwargs)
```

sdk/python/flet/page.py
```diff
@@ -294,6 +300,7 @@ def go(self, route):
             )
         )
         self.update()
+        self.query() # Update query url
```
FeodorFitsner, line 303: not sure I understand this line. Could you please elaborate?

sdk/python/flet/page.py
StanMathers: Hello, been busy recently due to university, so I was not able to refactor the code.

As for code, what I did is the following:
 I initialize `QueryString` object as the page is initialized. This case, every time route is changed, `QueryString` object must be updated to contain always up-to-date route. To contain current URL, I implemented __call__ method which is responsible for updating current URL inside of `QueryString` object every time `go` method is used.

As for `route_change` event, I consider it to be good use for updating current URL which I'll try soon.

As for page.query as property, I think that's a really good idea. I'll work on that soon.

sdk/python/flet/page.py
StanMathers: Certainly, I'll refactor this/

sdk/python/flet/page.py
StanMathers: Well, as I mentioned in previous comment, `self.query()` is using `__call__` magic method to keep updating url every time `go` method is called, but I'm thinking updating url, using `route_change` event.

FeodorFitsner: ah, magic! :)

sdk/python/flet/page.py
FeodorFitsner: yep. `route_change` is called when you change URL manually in the browser address bar.

FeodorFitsner:
Are regular expressions still used? Looks like there is some unused code:
https://github.com/flet-dev/flet/pull/280/files#diff-254a2814a7c09b3230e978750f3dab115e3fc59d1e243d1e62ceff2e051a5124R74-R75
https://github.com/flet-dev/flet/pull/280/files#diff-254a2814a7c09b3230e978750f3dab115e3fc59d1e243d1e62ceff2e051a5124R84

StanMathers:
> Are regular expressions still used? Looks like there is some unused code: https://github.com/flet-dev/flet/pull/280/files#diff-254a2814a7c09b3230e978750f3dab115e3fc59d1e243d1e62ceff2e051a5124R74-R75 https://github.com/flet-dev/flet/pull/280/files#diff-254a2814a7c09b3230e978750f3dab115e3fc59d1e243d1e62ceff2e051a5124R84

Done, I reduced regex as much as possible. I left only one line of regex for checking if URL is encoded.

sdk/python/flet/querystring.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _querystring_part(self, url_string: bool = False):
+        """
+        Function sliced url part and returns querystring part.\n
+        Use case: checking querystring part for encode, assiging decoded value
+        """
+        pattern = re.compile(r"\?[\w\D]+")
+        data = pattern.search(self.url)
+        return data if url_string is False else self.url[data.start() + 1 : data.end()]
+
```
FeodorFitsner, line 47: Couldn't be query string extracted with just `urllib.parse.urlparse(self.url).query`, without using `re`, or I'm getting it wrong?

sdk/python/.gitignore
```diff
@@ -135,4 +135,5 @@ __pypackages__/
 # bin
 flet/bin/
 
-/playground
\ No newline at end of file
+/playground
+hello.py
```
FeodorFitsner, line 139: Could you please remove `hello.py`?

StanMathers: There have been merge conflicts which I fixed but now there is some pdm problems. I am not sure what it is, but I'll try fixing it on Sunday.

PR #369, opened 2022-09-21
Modify some syntax

modaye:
1. Modify some syntax

2. Upgrade the black version in pre commit: solve known problems https://github.com/psf/black/issues/2964


FeodorFitsner:
Beautiful!

PR #400, opened 2022-09-27
Use False instead of None as `ink` default value

YouJiacheng:
Align with documentation
> `True` to produce ink ripples effect when user clicks the container. Default is `False`.

No three-valued logic is needed here, using `Optional[bool]` doesn't make sense.


FeodorFitsner:
I understand. The idea of optional boolean in case of control props is that the property value is just not included in JSON message at all when it's `None` while with `ink: bool = False` it's always included into message `ink="false"`. Not a big deal, but if you have thousands of containers it could be significant addition to the traffic :) Does it make sense?

YouJiacheng:
That make sense!

PR #427, opened 2022-10-01
make ink=True behavior consistent with ink=False

YouJiacheng:
Use onTapDown instead of onTap like ink = False
Also:
passing a dummy callback to onTap when onHover is True to enble the widget instead of
passing a ws.pageEventFromWeb callback to onTap and onLongPress to reduce websocket traffic

### DIFF
--- a/package/lib/src/controls/container.dart
+++ b/package/lib/src/controls/container.dart
@@ -113,16 +113,27 @@ class ContainerControl extends StatelessWidget {
             var ink = Ink(
                 decoration: boxDecor,
                 child: InkWell(
-                  onTap: onClick || onHover
-                      ? () {
+                  // Dummy callback to enable widget
+                  // see https://github.com/flutter/flutter/issues/50116#issuecomment-582047374
+                  // and https://github.com/flutter/flutter/blob/eed80afe2c641fb14b82a22279d2d78c19661787/packages/flutter/lib/src/material/ink_well.dart#L1125-L1129
+                  onTap: onHover
+                      ? () {}
+                      : null,
+                  onTapDown: onClick
+                      ? (details) {
                           debugPrint("Container ${control.id} clicked!");
                           ws.pageEventFromWeb(
                               eventTarget: control.id,
                               eventName: "click",
-                              eventData: "");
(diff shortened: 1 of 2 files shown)


FeodorFitsner:
Could you please elaborate what you changes about?

FeodorFitsner:
Is it about [this](https://github.com/flet-dev/website/pull/44#issuecomment-1264489618)?

YouJiacheng:
Yes. First, I use onTapDown to trigger onclick callback instead of onTap when ink = True, which is the same as ink = False case, so that onclick callback can receive position data when ink = True.
Second, I use a dummy callback for onTap to keep widget enabled when onHover is True. Before this PR, full functionality callbacks were used for onTap AND onLongPress just to keep widget enabled when onHover is True, which will send message through ws.

FeodorFitsner:
Cool, thank you!

FeodorFitsner:
That makes sense.

YouJiacheng:
I haven't tested this PR with flet because I dont know how to build flet client, I only made something similar in flutter playground.

FeodorFitsner:
OK, I found an issue with having both handler `onTapDown` and `onLongPress`. It's impossible to distinguish between `onTapDown` and `onLongPress`: it triggers either just `onTapDown` or, if you hold it longer `onTapDown` and then `onLongPress`.
With `onTap` it triggers either `onTap` or `onLongPress`.

FeodorFitsner:
I suggest explicitly implement `on_tap_down` event, with coordinates.

FeodorFitsner:
I'm going to do this in the next update. Going to be a breaking change though.

YouJiacheng:
It is intended (in flutter) that `onTapDown` and `onLongPress` is "indistinguishable" -- actually `onTapDown` will always be triggered (it even be fired for loser in gesture arena, except very short tap).

I suggest refactor the API to directly expose flutter's `GestureDetector` API, with same name and same "interface".

It might be even better to implement a flet `GestureDetector` as a replacement of this functionality of `Container`(leave current API as deprecated, or just remove them -- breaking change). I don't think that we should integrate so many functions into `Container` -- “Single Responsibility Principle”.

Issue #433, opened 2022-10-03
Flutter: Clean up warnings


sh1l0n:
**Description**

Some Flutter warnings appear when load the project in VSCode using the default linter

**Describe the results you received:**

![Screenshot 2022-10-03 at 22 03 22](https://user-images.githubusercontent.com/11474367/193668859-c89b4e90-127f-4c8a-8776-384febfebee6.png)

**Describe the results you expected:**
0 warnings 

**Flet version (`pip show flet`)**:

```
0.1.60
```

**Operating system**:
```
macOS
12.3.1
```

**Additional environment details:**
```bash
$ flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.3.2, on macOS 12.3.1 21E258 darwin-arm, locale
    en-ES)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] VS Code (version 1.71.2)
[✓] Connected device (2 available)
[✓] HTTP Host Availability

• No issues found!
```

FeodorFitsner:
It should be fixed now.

PR #438, opened 2022-10-03
Flutter: Clean up warnings

sh1l0n:
https://github.com/flet-dev/flet/issues/433


FeodorFitsner:
As you are changing `pubspec.lock` in both PRs would it make sense to combine this PR and https://github.com/flet-dev/flet/pull/439 into one?

sh1l0n:
> As you are changing `pubspec.lock` in both PRs would it make sense to combine this PR and #439 into one?

from my point of view, much better to do it in separated PRs. Small features better than big ones

Issue #434, opened 2022-10-03
Flutter: Fix dependencies version


sh1l0n:
**Please Describe The Problem To Be Solved**

Needs to fix the flutter dependencies to a specific version in order to ensure a stable version. If we left the ^ , the flutter pub get command will try to download the latest version of the package that could end in a runtime crash

**pubspec.yaml**
![Screenshot 2022-10-03 at 22 10 19](https://user-images.githubusercontent.com/11474367/193669754-275ecebc-8acd-4874-ada4-2c57fe70b7ce.png)

**(Optional): Suggest A Solution**
Remove ^ and list packages alphabetically

PR #439, opened 2022-10-03
Flutter: Fix packages to specific versions

sh1l0n:
https://github.com/flet-dev/flet/issues/434


FeodorFitsner:
It's not recommended to commit package's `pubspec.lock`: https://dart.dev/guides/libraries/private-files#pubspeclock

PR #443, opened 2022-10-04
fix: ensure correct version is fetched in dev mode

skeledrew:
Because the version file isn't replaced when the repo is cloned, the version info may go out of date, which can cause issues for local dev.

### DIFF
--- a/sdk/python/flet/version.py
+++ b/sdk/python/flet/version.py
@@ -1,24 +1,28 @@
 # this file will be replaced by CI
 
 import subprocess as sp
+from pkg_resources import parse_version
+
+from flet.utils import which
 
 
 def update_version():
-    in_repo = sp.run(
+    in_repo = which("git") and sp.run(
         ["git", "status"],
         capture_output=True,
-        text=True
+        text=True,
     ).stdout.startswith("On branch ")
 
     if in_repo:
         # NOTE: this may break if there is a tag name starting with
         #         "v" that isn't a version number
-        version = sp.run(
-            ["git tag |grep '^v' |sort -V |tail -n 1 |sed 's/v//g'"],
+        tags = sp.run(
+            ["git", "tag"],
             capture_output=True,
-            shell=True,
-            text=True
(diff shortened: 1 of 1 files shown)


FeodorFitsner:
Thanks for the PR!

A few notes:

1) It will fail on Windows as it doesn't have `grep`, etc.
2) It's not optimal to call `sp.run` on `version.py` module load. Let's create some `get_last_release_version` method doing that on demand.

skeledrew:
1. Ah yes, I only thought of Windows after I pushed. I'll update.
2. I thought about performance, and it takes under 0.7 secs on my system when I timed it with `time python -c 'import flet; print(flet.version.version)'` which I thought OK. But I'll try to think of something; I want to ensure `import flet; flet.version.version` continues to work as is.

FeodorFitsner:
Ha, 2) is a good point. `version.py` is being [updated/replaced during CI](https://github.com/flet-dev/flet/blob/main/.appveyor.yml#L436), so we have to think about how to deal with that.

Maybe something like could work:

```python
version = ""

if not version:
  # get version from Git
```

and CI would update `version.py` instead of full replace.

skeledrew:
Actually I chose this hack as I wanted to avoid touching (and maybe breaking) the CI config. The convention I've noticed over the years (and adopted) is that at release time, the version is first bumped in all relevant files (using a tool like [bump2version](https://github.com/c4urself/bump2version), this change is committed, and then the release is tagged.

FeodorFitsner:
Thanks! How would CI update step look like then?

skeledrew:
Meaning the command sequence? I don't configure CIs much, but I'm thinking something like:
```bash
...
git add ...
git commit -m "feat: implement some awesome stuff"
pytest
bump2version minor  # update pyproject.toml and version.py
git commit -am "chore: bump version 0.1.60 -> 0.2.0"
git tag -a "v0.2.0" -m "Adds cool stuff"
git push
```

PR #454, opened 2022-10-07
Fix controls setter for empty list input

YouJiacheng:
User might pass an empty list to the setter then modify the list in-place with another name/variable referencing the list and expect the controls' content will change as well.
Current `value or []` will surprisingly break this use-case. Given that in-place modification will work if user pass non-empty list, this behavior is not satisfactory.
Also -- Google python style guide says: Always use `is None` and `is not None` to check for a `None` value. https://google.github.io/styleguide/pyguide.html#2144-decision

### DIFF
--- a/sdk/python/flet/alert_dialog.py
+++ b/sdk/python/flet/alert_dialog.py
@@ -138,7 +138,7 @@ class AlertDialog(Control):
 
     @actions.setter
     def actions(self, value):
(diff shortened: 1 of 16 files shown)


ndonkoHenri:
https://peps.python.org/pep-0008/#programming-recommendations

PEP 8 recommends it too.

PR #490, opened 2022-10-17
making Control a cooperative class

ndvanforeest:
I used the observer pattern to propagate changes in a usercontrol object. For that purpose my widget derives from UserControl and another class, called Subject.  Calling super().__init__() in my class did not work because I had the wrong order in the class calls, i.e., class MyWidget(UserControl, Subject), rather than class MyWidget(Subject, UserControl). By including super().__init__() in the Control class such points of confusion are easy to prevent. Here is a further discussion:

https://rhettinger.wordpress.com/2011/05/26/super-considered-super/

### DIFF
--- a/sdk/python/flet/control.py
+++ b/sdk/python/flet/control.py
@@ -100,6 +100,7 @@ class Control:
         disabled: Optional[bool] = None,
         data: Any = None,
     ):
+        super().__init__()
         self.__page: Optional[Page] = None
         self.__attrs = {}
         self.__previous_children = []


FeodorFitsner:
Interesting. Is that 'Subject' class your own? What problem were you solving? Is there something from your solution you can share with the rest of Flet community?

ndvanforeest:
First, thanks for making flet. It's a great tool.

Here is the class that I used:
https://www.geeksforgeeks.org/observer-method-python-design-patterns/. It's
not my own.

When changing some widget, I wanted to update another widget that presented
just another view on the data. Exampe: in widget one I append a number to a
list, and then  widget two presents the sum of the list. This second widget
has to update itself if the list gets changed in widget one.

If I am not mistaken, Qt implements this with signal and slots. Some
general support for the observer pattern would certainly be a useful
extension in flet. However, it's not hard to implement oneself, in
particular if UserControl becomes a cooperative class :-)

FeodorFitsner:
Got it, thanks!

PR #503, opened 2022-10-20
fix: handle a few corner cases

skeledrew:
- Check if `git` is installed.

- Ensure proper working directory when used in another project.

- Use the version of the active branch.

- Improved error messages if tag retrieval fails.

### DIFF
--- a/sdk/python/flet/version.py
+++ b/sdk/python/flet/version.py
@@ -1,15 +1,23 @@
+"""Provide the current Flet version."""
+
+
+import os
 import subprocess as sp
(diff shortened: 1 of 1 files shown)


FeodorFitsner:
It doesn't work for me on Windows 11, I'm always getting hardcoded `0.1.60`.

Removing `which("git") and ` makes it work. Why?

FeodorFitsner:
Got it! It should check for `which("git.exe")` on Windows and for `which("git")` on other OSes.

Could you do a fix please?

PR #522, opened 2022-10-26
Added scrollBehavior dragDevices to allow drag scroll on web & desktop

Skquark:


FeodorFitsner:
Could you please explain what that fix does?

Skquark:
The fix allows you to scroll any scrollable component (mainly column & listview) on the Web build with the mouse pointer to drag down/up instead of only having to use the mouse scroll wheel or touchpad to scroll a view.  So basically it adds the drag scroll feature that is in the native apps on a phone/tablet into the web app, which also applies to web view in a phone browser.  It's a simple feature that should of been turned on by default because there's no reason you wouldn't want it on, and makes the web ui feel more like a native app ui.  It's also particularly needed for web browser since the arrow keys (along with page up/down) doesn't scroll a flutter app like you can with HTML, so it's a needed capability.
Hope I made my case for that addition well enough, but I think other people running Flet on the web browser would appreciate having that ability turned on.  If you're still on the fence about turning drag scroll gestures on by default, it could be added as an option, but it'd be much easier to just enable for all, no downside since it doesn't affect any other scrolling method negatively.  Thanks.

ndonkoHenri:
I think this should be an option and not a default behavior.

FeodorFitsner:
`MaterialScrollBehavior` is deprecated with a notice *Use ThemeData.useMaterial3 or override ScrollBehavior.buildOverscrollIndicator. This feature was deprecated after v2.13.0-0.0.pre.*. I also found this: https://docs.flutter.dev/release/breaking-changes/default-scroll-behavior-drag

Skquark:
Didn't notice MaterialScrollBehavior got depreciated, but still worked.  I just updated it with the new method making the class to include all the PointerDeviceKinds as dragDevices.
I understand if you'd rather make it as an option to set instead of default scroll behavior for everyone, even though I think it's practical enough to enable for all.  If you want it as settable option, I'd recommend something like page.enable_web_drag_scrolling = True
and if it's False, to just have scrollBehavior: enable_web_drag_scrolling ? DragScrollBehavior : none,
Either way.. In my web app, I'm constantly finding myself trying to drag the scroll view down with my mouse and disappointed when I have to use the mouse wheel or grabbing the scrollbar.  It feels much more natural to have the feature enabled, but that may not be for everyone, can't speak for the average user.

Skquark:
Another case scenario in support of enabling this feature is when I shrink the width of my web app, the Tabs are cut off on the right and I can't scroll the TabBar to get to the ones on the end, however it does work on the mobile to drag it to get to the hidden Tabs.  Makes a difference for functionality, not just an alternative to dragging the small scroll bar or mouse wheel which gets tiring on the long lists..

ndonkoHenri:
Do you mean this: https://api.flutter.dev/flutter/material/TabBar/isScrollable.html ?

Skquark:
It seemed that the TabBar was already set to Scrollable because on a mobile with Touch Gesture, the tabs were able to scroll sideways, but on desktop or web with Mouse Gesture you can't scroll tabs horizontally with the mouse wheel.  It'd work with that  PointerDeviceKind.mouse set on dragDevices as per my PR, in addition to being able to drag any other scroll controller with the mouse.

FeodorFitsner:
`scrollBehavior` is applied to a wrong "MaterialApp". The one that displays entire application is here: https://github.com/flet-dev/flet/blob/main/package/lib/src/controls/page.dart#L438

FeodorFitsner:
I'm looking at [this page](https://api.flutter.dev/flutter/widgets/ScrollBehavior/dragDevices.html) 

Could you please comment on *"Enabling this for [PointerDeviceKind.mouse](https://api.flutter.dev/flutter/dart-ui/PointerDeviceKind.html) will make it difficult or impossible to select text in scrollable containers and is not recommended."*?

Would it be another reason to have `scrollBehavior` optional/configurable?

ndonkoHenri:
So this means, Flutter itself doesn't make use of this Behavior.

FeodorFitsner:
I'm not sure what's going on with this PR? Is it mix with ImageViewer? Shall we close it and start over?

Skquark:
Opps, sorry I should have kept things in a separate branch, so it got mixed with the ImageViewer changes.  What was modified for the scrollBehavior is just in package/lib/src/flet_app_main.dart so you can scrap my PR and start over since it was simple enough.  The only thing missing from this is a way to turn it on & off with a flag to enable it, and I didn't know the best way to set that.  All this PR adds is this to the bottom of flet_app_main.dart:
```dart
class DragScrollBehavior extends MaterialScrollBehavior {
  @override
  Set<PointerDeviceKind> get dragDevices => {
        PointerDeviceKind.touch,
        PointerDeviceKind.mouse,
        PointerDeviceKind.stylus,
        PointerDeviceKind.unknown,
      };
}
```
And on the top adding `import 'package:flutter/gestures.dart';`
And on MaterialApp( to add `scrollBehavior: enable_drag_scrolling ? DragScrollBehavior() : null,`
But I don't know your preferred way to add a bool flag arg for enable_drag_scrolling... Maybe that's an attribute that can be set in page.py or elsewhere, I'm not an expert in that.  I'd appreciate if someone can work that little part out and implement this simple change. It's still a useful addition that I'm sure more folks than just me would like to enable on their desktop/web app.

FeodorFitsner:
Closing as stale. Re-open and resolve conflicts if decide to work on this PR further.

PR #554, opened 2022-11-08
Controls S3

FeodorFitsner:


sdk/python/flet/tooltip.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # prefere_below
+    @property
+    def prefere_below(self) -> Optional[bool]:
+        return self._get_attr("prefereBelow", data_type="bool", def_value=False)
+
+    @prefere_below.setter
+    @beartype
+    def prefere_below(self, value: Optional[bool]):
+        self._set_attr("prefereBelow", value)
```
ndonkoHenri, line 204: There is a little issue over here. I had to come right to this commit to see how the attribute was named in the `tooltip.dart`.
The little issue is the a simple grammar mistake: `prefer` and not `prefere`. (notice the `e` at the end)
Will make a PR for this right now.

PR #580, opened 2022-11-16
Adds ZERO_PADDING variable for the default padding value instead of None. (fixes #578)

### LINKED ISSUE #578: text disappear when set TextField height property and value of TextField  string length grow bigger
```
from time import sleep

import flet
from flet import Column, Page, ProgressBar, Text, TextField, icons, padding
from flet.progress_bar import ProgressBar
from flet.textfield import TextField


def main(page: Page):
    page.title = "TextField Examples"
    page.window_width = 400
    page.window_height = 400
    t1 = TextField(value="--------------------1---------------")
    t1.height = 40
    t2 = TextField(value="--------------------1---------------")
    page.add(t1)
    page.add(t2)

flet.app(target=main)

```

https://user-images.githubusercontent.com/4066126/202209625-2a03881a-3f15-4ba3-b1a5-286f2c0925d2.mp4

maybe is content_padding is too larger!

```
from time import sleep

import flet
from flet import Column, Page, ProgressBar, Text, TextField, icons, padding
from flet.progress_bar import ProgressBar
from flet.textfield import TextField


def main(page: Page):
    page.title = "TextField Examples"
    page.window_width = 400
    page.window_height = 400
    t1 = TextField(value="--------------------1---------------")
    t1.content_padding = 0
    t1.height = 40
    t2 = TextField(value="--------------------1---------------")
    page.add(t1)
    page.add(t2)

flet.app(target=main)


```

when set content_padding =0  is OK!

pouralijan:
I added a new variable to padding.py named ZERO_PADDING for the default value instead of None.

For margin can be similar to this PR. If you want to apply for the margin, I can create a new RP. I hope, I have could help with this project.

### DIFF
--- a/sdk/python/flet/alert_dialog.py
+++ b/sdk/python/flet/alert_dialog.py
@@ -6,6 +6,7 @@ from flet.buttons import OutlinedBorder
 from flet.control import Control, MainAxisAlignment
 from flet.ref import Ref
 from flet.types import PaddingValue
+from flet.padding import ZERO_PADDING
 
 
 class AlertDialog(Control):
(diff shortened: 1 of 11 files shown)


FeodorFitsner:
Oh, unfortunately, it doesn't work like that.

Setting `padding` to `None` leaves it to be default padding determined by control of a theme.
For example, default `View's, `Container's, `ListView`s, `AlertDialog`s padding is 10.

This PR, on the other hand, forces all paddings to be 0. Merging this PR would practically "shrink" many UI elements.

ndonkoHenri:
Can I ask why? 
Did you encounter an error with `None`?

pouralijan:
Oh, So for that issue (#578), I will change content_padding on textfield.py if you agree.

FeodorFitsner:
With `content_padding=0` TextField looks like:

image

image

PR #621, opened 2022-11-22
Ability to change border shape of FAB control

### LINKED ISSUE #602: Border Radius FloatingActionButton
Insert `border_radius` in the FloantingActionButton control as it is in the Container.

Currently it is not possible to change, to have a button with a customized border radius, I use the `Container` as a button, I think that would not be ideal.

FeodorFitsner:
Close #602

`FAB.mini` also added.


ndonkoHenri:
It seems pretty easy to expose new properties. 🤔
Only the dart part of it seems unpredictable(for non dart devs).

FeodorFitsner:
Told ya! :)

PR #697, opened 2022-12-09
Fix typing caused by replacement of Literal with Enums; typos; indentation etc

ndonkoHenri:
The changes in this PR will resolve the issues in the image below(mainly caused by the replacement of Literals with Enums):

image

Happy coding! 🙂


FeodorFitsner:
Thanks for your PR!

CI is failing: https://ci.appveyor.com/project/flet-dev/flet/builds/45636350 could you please check?
You can run `pdm run pytest tests` before check-in to catch these issues locally.

ndonkoHenri:
To be honest, I rarely have to deal with tests. 😅
I will try my best and get back to you asap.

ndonkoHenri:
Here we are!
This was definitely a 'Testing for Nerds' course.

ndonkoHenri:
Can you please check the fail with MacOS build?

FeodorFitsner:
Looking into that.

sdk/python/flet/alert_dialog.py
```diff
@@ -61,26 +61,27 @@ def open_dlg_modal(e):
...
-        title: Optional[Control] = None,
-        title_padding: PaddingValue = None,
-        content: Optional[Control] = None,
-        content_padding: PaddingValue = None,
-        actions: Optional[List[Control]] = None,
-        actions_padding: PaddingValue = None,
-        actions_alignment: MainAxisAlignment = MainAxisAlignment.NONE,
-        shape: Optional[OutlinedBorder] = None,
-        on_dismiss=None,
+            self,
```
FeodorFitsner, line 66: Something has happened to an indentation here. Could you please check/fix?

sdk/python/flet/animated_switcher.py
```diff
@@ -112,9 +113,9 @@ def __init__(
         #
         duration: Optional[int] = None,
         reverse_duration: Optional[int] = None,
-        switch_in_curve: Optional[AnimationCurve] = None,
-        switch_out_curve: Optional[AnimationCurve] = None,
-        transition: Optional[AnimatedSwitcherTransition] = None,
+        switch_in_curve: Union[AnimationCurve, AnimationCurveString] = None,
```
FeodorFitsner, line 116: It's not an optional anymore?

sdk/python/flet/alert_dialog.py
ndonkoHenri: Fixed in next commit.

sdk/python/flet/animated_switcher.py
ndonkoHenri: If you want I can change it to:  `Union[AnimationCurve, AnimationCurveString, None]`
But then, I already added an Optional at the level of the function to deal with that..
```python
@beartype
    def __set_switch_in_curve(self, value: Optional[AnimationCurveString]):
```
What should I do please?

FeodorFitsner: Yep, I think `Union[AnimationCurve, AnimationCurveString, None]` is the way to go.

Otherwise it shows an error:

image

FeodorFitsner: Check other similar cases too if there are any.

FeodorFitsner: or just add `None` to `AnimationCurveString`?

sdk/python/flet/container.py
```diff
@@ -100,16 +101,16 @@ def __init__(
...
         image_src: Optional[str] = None,
         image_src_base64: Optional[str] = None,
-        image_repeat: Optional[ImageRepeat] = None,
-        image_fit: Optional[ImageFit] = None,
+        image_repeat: Union[ImageRepeat, ImageRepeatString] = None,
+        image_fit: Union[ImageFit, ImageFitString] = None,
         image_opacity: OptionalNumber = None,
-        shape: Optional[BoxShape] = None,
-        clip_behavior: Optional[ClipBehavior] = None,
+        shape: Union[BoxShape, BoxShapeString] = None,
```
FeodorFitsner, line 112: This one shows error - add `None` to Union or `BoxShapeString`.

sdk/python/flet/datatable.py
```diff
@@ -252,12 +253,12 @@ def cells(self, value):
```
FeodorFitsner, line 253: Check `DataTable` - all color properties are errored.

sdk/python/flet/gesture_detector.py
```diff
@@ -106,70 +118,71 @@ def on_pan_update2(e: ft.DragUpdateEvent):
...
-        on_pan_update=None,
-        on_pan_end=None,
-        on_scale_start=None,
-        on_scale_update=None,
-        on_scale_end=None,
-        on_hover=None,
-        on_enter=None,
-        on_exit=None,
-        on_scroll=None,
+            self,
```
FeodorFitsner, line 123: Indentation issue.

sdk/python/flet/animated_switcher.py
ndonkoHenri: added `None` to `Union`

sdk/python/flet/datatable.py
ndonkoHenri: Oh. Whats the error message?
Here is an example of what I used: `heading_row_color: Union[None, str, Dict[Union[MaterialState, MaterialStateString], str]]`
Any error?

ndonkoHenri:
By the way, don't you think the contents of the `DataTable`'s color props should change? For example(pay attention to line 4 and 5 of the block below) and assume the developer gave in a `Dict[MaterialState, str]` (MaterialState as key) as value:

```python
@heading_row_color.setter
    def heading_row_color(self, value: Union[None, str, Dict[Union[MaterialState, MaterialStateString], str]]):
        self.__heading_row_color = value
        if isinstance(value, MaterialState):
            self._set_attr("headingRowColor", value.value)
        else:
            self.__set_data_row_color(value)

    @beartype
    def __set_heading_row_color(self, value: Union[None, str, Dict[Union[MaterialState, MaterialStateString], str]]):
        self._set_attr("headingRowColor", value)
```

You are still using just `value.value` , but note that this is different! To reach the `MaterialState` Enum we pass through the `Dict`. So you instead have to be checking if the items of the `Dict` are instances of the `MaterialState` Enum and not the `Dict` itself.

sdk/python/flet/datatable.py
```diff
@@ -252,12 +253,12 @@ def cells(self, value):
 
     # color
     @property
```
FeodorFitsner, line 255: `DataRow.color` shows error:

image

sdk/python/flet/dropdown.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         label: Optional[str] = None,
         label_style: Optional[TextStyle] = None,
         icon: Optional[str] = None,
-        border: Optional[InputBorder] = None,
+        border: Union[InputBorder, InputBorderString] = None,
```
FeodorFitsner, line 85: Please check `FormField.border` as well.

sdk/python/flet/querystring.py
```diff
@@ -97,8 +97,8 @@ def __call__(self):
         # Checking if self.url is encoded and decoding it accordingly
         if self._is_encoded() is True:
             self.url = (
-                self.page.url
-                + urllib.parse.urlparse(self.url).path
-                + "?"
-                + self._decode_url_component(self._querystring_part(url_string=True))
+                    self.page.url
```
FeodorFitsner, line 100: Please fix indentation.

sdk/python/flet/row.py
```diff
@@ -61,43 +61,43 @@ def main(page: ft.Page):
...
-        #
-        alignment: MainAxisAlignment = MainAxisAlignment.NONE,
-        vertical_alignment: CrossAxisAlignment = CrossAxisAlignment.NONE,
-        spacing: OptionalNumber = None,
-        tight: Optional[bool] = None,
-        wrap: Optional[bool] = None,
-        run_spacing: OptionalNumber = None,
-        scroll: Optional[ScrollMode] = None,
-        auto_scroll: Optional[bool] = None,
+            self,
```
FeodorFitsner, line 64: Indentation.

ndonkoHenri:
🙂

ndonkoHenri:
@FeodorFitsner please I am waiting on this. I get typing errors everywhere...
It gives me the feeling I am doing stuffs the wrong way.

FeodorFitsner:
No worries, I'm on it.

sdk/python/flet/tooltip.py
```diff
@@ -84,10 +85,10 @@ def __init__(
         gradient: Optional[Gradient] = None,
         border: Optional[Border] = None,
         border_radius: BorderRadiusValue = None,
-        shape: Optional[BoxShape] = None,
+        shape: Union[BoxShape, BoxShapeString] = None,
```
FeodorFitsner, line 88: This shows error:

image

sdk/python/flet/tooltip.py
```diff
@@ -84,10 +85,10 @@ def __init__(
...
         message: Optional[str] = None,
         text_style: Optional[TextStyle] = None,
-        text_align: TextAlign = TextAlign.NONE,
+        text_align: Union[TextAlign, TextAlignString] = TextAlign.NONE,
```
FeodorFitsner, line 91: Let's make it `None` as everything else?

sdk/python/flet/text.py
```diff
@@ -130,14 +130,14 @@ def __init__(
...
         #
-        text_align: TextAlign = TextAlign.NONE,
+        text_align: Union[TextAlign, TextAlignString] = TextAlign.NONE,
         font_family: Optional[str] = None,
         size: OptionalNumber = None,
-        weight: Optional[FontWeight] = None,
+        weight: Union[FontWeight, FontWeightString] = None,
         italic: Optional[bool] = None,
-        style: Optional[TextThemeStyle] = None,
+        style: Union[TextThemeStyle, TextThemeStyleString] = None,
```
FeodorFitsner, line 138: `style` is underlined as an error here.

FeodorFitsner:
You did a huge job here - thanks!

But I'm thinking... would it be more explicit and correct to write instead of:

```
Union[TextOverflow, TextOverflowString]
```

with explicit:

```
Union[None, TextOverflow, TextOverflowString]
```

Because right now `None` is a part of `TextOverflowString` and it's kind of hard to see at a glance. With a latter notation a user immediately see that it could be either `None`, `TextOverflow` enum or `TextOverflowString` string. What do you think? I mean do all unions like that. If you think I'm asking to much - no worries, I can do that when this PR is merged. But I thought while you are on it... :)

FeodorFitsner:
like you did in https://github.com/flet-dev/flet/blob/db259f593f6a7a6f1615f24506f419a23b6b3f50/sdk/python/flet/animated_switcher.py#L116

FeodorFitsner:
Also, the main question is how long are we going to maintain two versions of those enums and string literals in each case? My original idea was switching to enums all together, rewrite all current examples to use enums and let developers see type errors when strings are used, but not enums. Are we going to have double enum-strings for new controls? Probably not.

So, maybe "fixing" that doesn't make sense at all?

FeodorFitsner:
It kind of confuses new developers too: should I use enum or a string? Right?

sdk/python/flet/tooltip.py
ndonkoHenri: I added `None` to the `TextAlignString` Literal.
Is that what you meant?

ndonkoHenri:
I get your point, and I think you are right. creating string and Enums for every prop is certainly not clean code and efficient! Thanks for the feedback.
Just that, seeing my code stained errors (typing), makes me feel I inserted the wrong value, or that I am doing stuffs the wrong way.
image

I personally prefer using strings instead of Enums, for the properties I already know.
Anyway, let's keep my preferences aside and do it your way, -- close this pr..?

FeodorFitsner:
Yeah, we'd love to make the transition to enums as smooth as possible, but it, unfortunately, leads to a code duplication. But it's 0.x version anyway which says the API could be changed multiple times 🤷‍♂️, right?

Let's close this PR, sorry 😥. You are doing a lot for this project and I really appreciate that! I'm going to create a new Discord channel where we could discuss any changes, or just DM me.

Thank you!

ndonkoHenri:
> I'm going to create a new Discord channel where we could discuss any changes
Alright.
Thanks!

sdk/python/flet/container.py
```diff
@@ -359,29 +360,36 @@ def content(self, value: Optional[Control]):
...
+        if isinstance(value, BoxShape):
+            self._set_attr("shape", value.value)
+        else:
+            self.__set_shape(value)
+
     @beartype
-    def shape(self, value: Optional[BoxShape]):
-        self._set_attr("shape", value.value if value is not None else None)
+    def __set_shape(self, value: BoxShapeString):
+        self._set_attr("shape", value)
```
ndonkoHenri, line 376: I am just from noticing the importance of this part, and similar parts of this PR.
image
This should _not_ be a normal behavior @FeodorFitsner. One should be able to pass either a string or a `BoxShape `. At the moment, only a `BoxShape` is accepted, while strings give an error as shown above.
What do you think?

ndonkoHenri: Hey @FeodorFitsner

PR #745, opened 2022-12-22
python 3 changes

marksmayo:
- added trailing commas (python 3.5, 3.6+)
 - tidied whitespace
 - added f-strings where I could
 - fixed comparisons to None
 - cleaned some imports
 - changed exit() to sys.exit()
 - tidied some other small pylint errors


FeodorFitsner:
Thanks! I'd like to import functional and cleanup changes, but formatting... Flet uses Black formatting and we wouldn't like to change that at the moment. I really don't want to focus on formatting right now.

PR #748, opened 2022-12-22
Create datetime_field.py

qraxiss:
I'm added dropdown & row based date time input field. 
More info: github.com/qraxiss/CalendarFlet

### DIFF
--- a/sdk/python/flet/datetime_field.py
+++ b/sdk/python/flet/datetime_field.py
@@ -1,0 +1,117 @@
+import datetime as dt
+import flet as ft
+import calendar
+
+from typing import Any, Optional, Union, Iterable, Callable
+
+from beartype.typing import List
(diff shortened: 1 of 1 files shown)


ndonkoHenri:
Nice!
You could instead try adding this to the [flet-dev/examples](https://github.com/flet-dev/examples) repo. 
Honestly, I don't think it will be quite useful here.

This is just my humble opinion though.

qraxiss:
I am aware that my code is not very library-friendly. It does not exactly require a Flutter component, because rather needs an `ft.Page` object to work. I just made it for my own development. I guess it would be more appropriate to pull it into the examples section.

FeodorFitsner:
Yep, could you please make a PR to https://github.com/flet-dev/examples?

PR #789, opened 2023-01-02
Page configuration in a separate class.

StanMathers:
I've implemented a meta classes and a new method for handling basic page configuration for different purposes.

Here is a quick example:
```python
import flet as ft
from flet import BaseConfig

class WindowsWindowConfig(BaseConfig):
    theme=ft.Theme(color_scheme_seed='green')
    on_resize='on_resize'
    window_minimizable=False

class AndroidWindowConfig(BaseConfig):
    window_height=600
    window_width=400


def main(page: ft.Page):
    if page.platform == 'windows':
        page.from_config(WindowsWindowConfig)
    elif page.platform == 'android':
        page.from_config(AndroidWindowConfig)
    
    page.add(ft.Text("Hello Word"))
    page.add(ft.Text(page.platform))

ft.app(target=main)
```
# Details
## New file and a method

- There is a new Python file named `metaconfig.py` which consists of a meta class and a class inheriting from the metaclass. Those are `Meta` and `BaseConfig`
- There is a new method in the `Page` class named `from_config(config: Type[BaseConfig])` which takes a `class that inherits from BaseConfig` and has page attributes

## How to use
As shown in the example above, you
- At first, must create a class and inherit from `BaseConfig`
- You can include any properties, method or events from [Page docs](https://flet.dev/docs/controls/page), there are no hardcoded values
- At last, pass your class as an argument to `page.from_config()` method and you're good to go

## Use cases
- You may want to have different page configurations for different platforms
- You may want to use different page configuration for a production and a development server which you can define beforehand
- You may want to have configuration settings in a separate file to maintain a clean code
- While testing, you may need to deal with different configuration modes and you don't need to rewrite everything from scratch

## Note:
You can also have methods inside of those classes as far as they have `@staticmethod` decorator and they don't have to access a `page` object directly.

Methods must be passed as `string` to variables

```python
class WindowsWindowConfig(BaseConfig):
    theme=ft.Theme(color_scheme_seed='green')
    on_resize='on_resize'
    window_minimizable=False
    on_resize="on_resize"
    
    @staticmethod
    def on_resize(e):
        print("Page has been resized")

```

### DIFF
--- a/sdk/python/flet/__init__.py
+++ b/sdk/python/flet/__init__.py
@@ -151,3 +151,4 @@ from flet.user_control import UserControl
 from flet.vertical_divider import VerticalDivider
 from flet.view import View
 from flet.window_drag_area import WindowDragArea
+from flet.metaconfig import BaseConfig
--- a/sdk/python/flet/metaconfig.py
+++ b/sdk/python/flet/metaconfig.py
@@ -1,0 +1,29 @@
+from typing import List, Any, Dict, Tuple
+
+class Meta(type):
+
+    _exclude_attrs: List[str] = ['__module__', '__qualname__']
+    _include_attrs: Dict[Dict[str, Any], Any] = {}
+    
+    def __new__(cls, name: str, bases: Tuple[type], attrs: Dict[str, Any]) -> type:
+        
+        for i in attrs:
+            if i not in cls._exclude_attrs:
+                cls._include_attrs[name] = {i: attrs[i]}
+
+        return super().__new__(cls, name, bases, attrs)
+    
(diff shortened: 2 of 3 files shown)


FeodorFitsner:
Thank you for your PR!

I'm thinking, instead of adding this functionality to a core library wouldn't be more "flexible" to just extend `Page` class in a user program, e.g.:

```python
ft.Page.from_config = from_config
```

or you can create a Python module for extending Page class with additional methods?

FeodorFitsner:
What do you think?

StanMathers:
Well, you're right.

Instead of extending `Page` class, I thought it would be more flexible and clear to have third party package. It would take only one line of code to change into that.

PR #838, opened 2023-01-09
Replace Fletd server for desktop apps with a light-weight Python shim

FeodorFitsner:
This PR is the implementation of a [new Flet desktop architecture](https://flet.dev/blog/flet-mobile-update#flet-new-desktop-architecture).

It replaces Fletd server written in Go with a light-weight shim written in Python which results to:
1) Only 2 system processes to run Flet app: Python interpreter and Flutter client.
2) Less communication overhead (minus two network hops between Python and Fletd) and lower latency (shim uses TCP on Windows and Unix domain sockets on macOS/Linux).
3) Shim binds to 127.0.0.1 on Windows by default which is more secure by default.
4) The size of a standalone app bundle produced by `flet pack` reduced by ~8 MB.

The implementation paves the way to support Pyodide, iOS and Android.

Other notable changes:

* All controls loading resources from web URLs are able to load from local files, by providing a full path within the file system, and from `assets` directory by providing relative path: `Image.src`, `Audio.src`, `Page.fonts`, `Container.image_src`.
* `flet` Python package has separated into two packages: `flet-core` and `flet`.
* switched from PDM to Poetry.
* `beartype` removed everywhere.
* Faster mouse wheel scrolling on Windows.


hololeo:
"Sorry GO" 

Sorry GO
You got to go
You were strong and reliable. 
And faithful and dependable

So Sorry GO
Sad to see you go

FeodorFitsner:
It's still there with us though, as Fletd is needed for web apps.

But we could rewrite that in Python either (FastAPI?).

PR #839, opened 2023-01-09
feat: Add ref property to Control

iron3oxide:
This enables us to set the ref after creation and thus e.g. create ui components in a completely separate module than the one where they are connected to business logic.

### DIFF
--- a/sdk/python/flet/control.py
+++ b/sdk/python/flet/control.py
@@ -50,9 +50,7 @@ class Control:
         self.__data: Any = None
         self.data = data
         self.__event_handlers = {}
-        self.__ref = ref
-        if self.__ref:
-            self.__ref.current = self
+        self.ref = ref
 
     def _is_isolated(self):
         return False
--- a/sdk/python/flet/control.py
+++ b/sdk/python/flet/control.py
@@ -50,8 +50,9 @@ class Control:
         self.__data: Any = None
         self.data = data
         self.__event_handlers = {}
-        if ref:
-            ref.current = self
+        self.__ref = ref
+        if self.__ref:
+            self.__ref.current = self
 
     def _is_isolated(self):
         return False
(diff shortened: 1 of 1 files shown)


sdk/python/flet/control.py
```diff
@@ -50,8 +50,9 @@ def __init__(
         self.__data: Any = None
         self.data = data
         self.__event_handlers = {}
-        if ref:
-            ref.current = self
+        self.__ref = ref
+        if self.__ref:
+            self.__ref.current = self
```
FeodorFitsner, line 56: It's unnecessary to duplicate `ref` setter logic in the constructor. You could just remove:

```
        self.__ref = ref
        if self.__ref:
            self.__ref.current = self
```

and call property setter right away:

```
self.ref = ref
```

So, why would we need to keep the last used reference to a reference inside control: `self.__ref`?

FeodorFitsner: Maybe you could provide a sample of how you are going to use that?

iron3oxide: Let's say you wanted to reuse a component inside your application, but you need to use it with a ref because of your architecture or for some other reason. With this change, you could create the component in a `components/` module and simply import it wherever you connect it to the respective ref. Is that understandable or should I provide an example with code?

FeodorFitsner: 🤯 Please! Example wouldn't hurt! :)

iron3oxide: https://github.com/iron3oxide/flet-mvp-counter

Since the example had to be a bit more complex I simply modified my MVP counter example. You can find the most relevant sections in view.py and the components module.

iron3oxide: I just realized that I could also pass the ref to the function/method that returns the control and use it there. If there's anything that would make adding the ref property as proposed here impractical, I'd be fine doing that instead.

FeodorFitsner: If it works for you then I'd better close this PR. Honestly, I don't see a point of keeping "last used" reference inside a control instance.

PR #889, opened 2023-01-16
Add transparancy to matplotlib

AM1010101:
added property to set transparency of the matplotlib plots
implemented with simple if/else
closed previous pr to update property name to "transparent" instead of "transparency"

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/matplotlib_chart.py
+++ b/sdk/python/packages/flet-core/src/flet_core/matplotlib_chart.py
@@ -92,7 +92,7 @@ class MatplotlibChart(Container):
         #
         isolated: bool = False,
         original_size: bool = False,
-        transparency: bool = False,
+        transparent: bool = False,
     ):
 
         Container.__init__(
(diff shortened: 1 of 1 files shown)


AM1010101:
For some background this allows charts to appear like this:
image

instead of like this:
image

AM1010101:
or this:
image

sdk/python/packages/flet-core/src/flet_core/matplotlib_chart.py
```diff
@@ -133,7 +135,12 @@ def _before_build_command(self):
         super()._before_build_command()
         if self.__figure is not None:
             s = io.StringIO()
-            self.__figure.savefig(s, format="svg")
+
+            if self.__transparent:
+                self.__figure.savefig(s, format="svg", transparent=True)
+            else:
+                self.__figure.savefig(s, format="svg")
+
```
FeodorFitsner, line 143: Why not just `self.__figure.savefig(s, format="svg", transparent=self.__transparent)`?

AM1010101: I would agree that's better

PR #892, opened 2023-01-16
Add transparancy to matplotlib

AM1010101:


FeodorFitsner:
Dup https://github.com/flet-dev/flet/pull/889

PR #904, opened 2023-01-19
Add default values in Padding, Margin, Offset, BorderRadius dataclass

zrr1999:
Similar to https://github.com/flet-dev/flet/pull/883


FeodorFitsner:
I don't remember exactly where, but we were discussing this problem before. Enforcing zeros instead of `None` as default values for margins, paddings, offsets, etc. will break default built-in values of individual controls.

PR #925, opened 2023-01-23
New DateField field control

berecik:
A button lets the user select date on datepicker dialog.

Example:
```
import flet as ft

def main(page: ft.Page):
    def button_clicked(e):
        t.value = f"Datetime value are:  '{dt.value}'."
        page.update()

    t = ft.Text()
    dt = ft.DateField(value=DateTime.now(), on_change=button_clicked)
    b = ft.ElevatedButton(text="Submit", on_click=button_clicked)
    page.add(dt, b, t)

ft.app(target=main)
```

-----


ndonkoHenri:
Please for an image or screen record of your control in action. 🙂

FeodorFitsner:
Thank you for your submission!

Some feedback:

1. I see you are using an `ElevatedButton` as a control opening a date selection dialog, but you have inherited `DateField` Python counterpart from a `TextField`. I would do that component as a non-visual and renamed to `DatePicker`. So, it has  some "pick_date()` method and then "on_result" event, similar to [FilePicker](https://github.com/flet-dev/flet/blob/main/sdk/python/packages/flet-core/src/flet_core/file_picker.py) control. This way we are not dictating the way selected date is displayed to the user. Does it make sense?

FeodorFitsner:
See DatePicker in Flutter's Gallery app: https://gallery.flutter.dev/#/demo/pickers

package/pubspec.yaml
```diff
@@ -40,6 +40,10 @@ dependencies:
   shake: ^2.1.0
   path: ^1.8.2
 
+  # DatePicker specyfic dependencies
+#  flutter_datetime_picker:
+  intl:
```
FeodorFitsner, line 45: put a version for `intl` package and revert other changes.

berecik:
> 

I absolutely agree. My problem is, I need a form field, not just a picker. So this is a reason to use a name datefield like textfield, etc. but not a date_picker, which will suggest similar to file_picker functionality. I thought about doing a split for GenericField and DatePicker as the next step. But with your suggestion, I'll do this now.

FeodorFitsner:
Closing in favour of https://github.com/flet-dev/flet/pull/930

Issue #935, opened 2023-01-27
PaddingValue is not exposed


iron3oxide:
**Description**

When subclassing flet.View, one has to collect and then pass a lot of parameters to the __init__() method of it in order to not lose any functionality. Thankfully, VSCode is smart enough to create such a statement automatically and even add the required type hints. 

Two of the type hints cause problems though: `OptionalNumber` and `PaddingValue` are both not found in the flet package. Weirdly enough, the former should be there as it is imported in the `__init__.py` of flet_core and the `__init__.py` of flet simply imports everything from flet_core. The latter seems to have been forgotten though.

**Screenshot**
![image](https://user-images.githubusercontent.com/98779754/215220101-2aefb501-66dc-4db7-8a09-8b5a686c2b6b.png)

**Flet version (`pip show flet`)**:

0.4.0.dev1117

iron3oxide:
In an even weirder turn of events, `OptionalNumber` is now suddenly recognized as part of flet. Will make a PR to include `PaddingValue` as well and then close this.

PR #936, opened 2023-01-27
feat: Add PaddingValue to __init__.py

iron3oxide:
See #935


FeodorFitsner:
What about `OptionalNumber`?

iron3oxide:
It seems to be included in the 4.0.0 prereleases (not in 3.2.0 though), so I felt no need to add it.

FeodorFitsner:
Alright!

PR #948, opened 2023-01-28
Date picker control

berecik:
Complete Date Picker control, ready to use, with tests.


package/lib/src/controls/create_control.dart
```diff
@@ -220,6 +221,12 @@ Widget createControl(Control? parent, String id, bool parentDisabled) {
               control: controlView.control,
               children: controlView.children,
               parentDisabled: parentDisabled);
+        case "date_picker":
```
FeodorFitsner, line 224: Let's use the same naming convention for control names as the rest - lowecase and without `_` or `-` inside.

package/lib/src/controls/date_picker.dart
```diff
@@ -0,0 +1,171 @@
...
+              FletAppServices.of(context).server.sendPageEvent(
+                  eventTarget: widget.control.id,
+                  eventName: "change",
+                  eventData: stringValue);
+            }
+            FletAppServices.of(context).server.sendPageEvent(
+                eventTarget: widget.control.id,
+                eventName: "submit",
+                eventData: "");
+          }
```
FeodorFitsner, line 105: What's the difference between `change` and `submit` events?

package/lib/src/controls/date_picker.dart
```diff
@@ -0,0 +1,171 @@
...
+
+          void onChanged(DateTime? dateValue) {
+            debugPrint("New date: $dateValue");
+            var newState = "initState";
+            setState(() {
+              _value = dateValue;
+              _state = newState;
+            });
+            String stringValue = dateValue?.toIso8601String() ?? "null";
```
FeodorFitsner, line 88: Use empty string (`""`) instead of `"null"`.

sdk/python/packages/flet-core/src/flet_core/date_picker.py
```diff
@@ -0,0 +1,315 @@
...
+
+    # value
+    @property
+    def value(self) -> Optional[datetime]:
+        value_string = self._get_attr("value", def_value=None)
+        if value_string is None or value_string == 'null':
+            return None
+        else:
+            return datetime.fromisoformat(value_string)
+
```
FeodorFitsner, line 174: Now here if it's `None` or an empty string you can do just:

```python
return datetime.fromisoformat(value_string) if value_string else None
```

package/lib/src/controls/date_picker.dart
```diff
@@ -0,0 +1,171 @@
...
+          if (localeString == null) {
+            locale = Localizations.localeOf(context);
+          } else {
+            locale = Locale(localeString);
+          }
+
+          void onChanged(DateTime? dateValue) {
+            debugPrint("New date: $dateValue");
+            var newState = "initState";
+            setState(() {
```
FeodorFitsner, line 84: `setState` is unnecessary in Flet controls and they are rebuilt on `Store` changes. Also, `setState` is not needed to persist a value withing a widget as all widgets are basically stateful objects already.

package/lib/src/controls/date_picker.dart
berecik: change is on any text field change, submit is on select choice

FeodorFitsner: yeah, but do both events are triggered all the time when `onChange` is true?

FeodorFitsner: In [the docs](https://api.flutter.dev/flutter/material/showDatePicker.html) I see you can only grab selected date value in the future returned by `showDatePicker()`. Looks like we can implement `submit` only, right?

berecik: yes, it was an artefact from old version

berecik: now is on_change only available and working

FeodorFitsner: have you pushed the latest? I still see "submit" in both dart and python parts.

FeodorFitsner:
@berecik Are you still working on this PR?

FeodorFitsner:
Close in favor of #1998

PR #971, opened 2023-02-03
Changed re-connection logic to make hot reload work

FeodorFitsner:


ndonkoHenri:
Can you please briefly explain what the issue was?

FeodorFitsner:
Reconnecting logic was not working in TCP client.

FeodorFitsner:
Because we need more testing 🤷🏻‍♂️

PR #1066, opened 2023-02-20
added `star-history`

hemangjoshi37a:
added `star-history`

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,3 +117,8 @@ run again and now you instantly get a web app:
 ## Contribute to this wonderful project
 
 * Read the <a href="https://github.com/flet-dev/flet/blob/main/CONTRIBUTING.md">CONTRIBUTING.md</a> file
+
+## Star History
+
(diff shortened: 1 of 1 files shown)


hemangjoshi37a:
Also please take a look at my issue : #1067

FeodorFitsner:
I, honestly, a little hesitate to put that chart into the readme :) Such charts are good when starts are growing. But what would we do in case the growth slows down - remove the chart?

hemangjoshi37a:
> I, honestly, a little hesitate to put that chart into the readme :) Such charts are good when starts are growing. But what would we do in case the growth slows down - remove the chart?

It's not about showing the growth. its about being real and authentic about how things were and how things are now. It might be possible that if the growth slows down someone might take interest to revive the project through PRs.

hemangjoshi37a:
> I, honestly, a little hesitate to put that chart into the readme :) Such charts are good when starts are growing. But what would we do in case the growth slows down - remove the chart?

Regardless it's just cool. 🆒 😎

PR #1070, opened 2023-02-21
Add template for Q&A discussions

iron3oxide:
Since discussions within that category are mostly help requests and I frequently find myself having to ask for more info from the OP, e.g. an error message or a code sample, I thought it would be nice to have a template. It can be seen in action on this fork.

The GitHub discussion category form feature is in public beta, so this template might suddenly not work anymore because GitHub changed the syntax or something else broke. It is my understanding though that in that case the default template would simply be chosen, so it wouldn't be a major inconvenience.

Apologies that it's 15 commits when it could have been 1 or 2, it was a trial and error process and much to my disdain, GitHub didn't document what the category slug for categories containing an "&" is.


FeodorFitsner:
Cool! I see it's being triggered when `help wanted` label is applied to a discussion - it works on your fork. Could you tell how to configure the same in Flet's repo?

iron3oxide:
It works a bit differently: it gets triggered when you start a Q&A discussion, presents the user with a custom form and automatically adds the "help wanted" label (the user can of course remove that before or after posting). You can try that out in the fork.

There should be no configuration necessary, when a discussion is started, GitHub will automatically look for a yaml file in the `.github/DISCUSSION_TEMPLATE/` directory that is named after the slug of the category the discussion is intended to be started in, in this case `q-a.yml`. One can therefore easily add templates for other categories, see the docs [here](https://docs.github.com/en/discussions/managing-discussions-for-your-community/creating-discussion-category-forms).

.github/DISCUSSION_TEMPLATE/q-a.yml
```diff
@@ -0,0 +1,34 @@
+title: "[Question]"
+labels: ["help wanted"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This is the place to ask for specifics about flet or for help with problems you encountered while using it.
```
FeodorFitsner, line 7: Could you change `flet` to `Flet` please?

iron3oxide: done

PR #1076, opened 2023-02-22
Added `Container.blur` and `Container.shadow` properties

FeodorFitsner:
Blur docs: https://blog.logrocket.com/enhance-mobile-apps-with-flutter-boxshadow/
Shadow examples: https://blog.logrocket.com/enhance-mobile-apps-with-flutter-boxshadow/


sdk/python/packages/flet-core/src/flet_core/colors.py
```diff
@@ -356,3 +356,14 @@
 DEEP_ORANGE_ACCENT_200 = "deeporangeaccent200"
 DEEP_ORANGE_ACCENT_400 = "deeporangeaccent400"
 DEEP_ORANGE_ACCENT_700 = "deeporangeaccent700"
+GREY = "grey"
```
ndonkoHenri, line 359: Please let me know when you add colors. (for my colors browser)
I added this "grey" color right now.
Did flutter 3.7 bring in some other new ones?

sdk/python/packages/flet-core/src/flet_core/container.py
```diff
@@ -35,6 +38,36 @@
...
+    tile_mode: BlurTileMode = field(default=BlurTileMode.CLAMP)
+
+
+@dataclasses.dataclass
+class BoxShadow:
+    spread_radius: Optional[float] = field(default=None)
+    blur_radius: Optional[float] = field(default=None)
+    color: Optional[str] = field(default=None)
+    offset: OffsetValue = field(default=None)
+    tile_mode: ShadowBlurStyle = field(default=ShadowBlurStyle.NORMAL)
```
ndonkoHenri, line 68: Inconsistent naming here. `blur_style` could be a better and consistent name. 
What do you think?

FeodorFitsner: Good catch, thank you! Will fix that shortly!

sdk/python/packages/flet-core/src/flet_core/colors.py
FeodorFitsner: I don't think it was introduced in 3.7 - looks like a missed it in the first place ☺️

sdk/python/packages/flet-core/src/flet_core/container.py
```diff
@@ -35,6 +38,36 @@
...
+    NORMAL = "normal"
+    SOLID = "solid"
+    OUTER = "outer"
+    INNER = "inner"
+
+
+@dataclasses.dataclass
+class Blur:
+    sigma_x: float
+    sigma_y: float
```
ndonkoHenri, line 58: Why only `float`? 
`Union[int, float]` won't be better?

sdk/python/packages/flet-core/src/flet_core/container.py
```diff
@@ -35,6 +38,36 @@
...
+    tile_mode: BlurTileMode = field(default=BlurTileMode.CLAMP)
+
+
+@dataclasses.dataclass
+class BoxShadow:
+    spread_radius: Optional[float] = field(default=None)
+    blur_radius: Optional[float] = field(default=None)
```
ndonkoHenri, line 65: Same here. `Union[int, float]`

Issue #1073, opened 2023-02-21
Transforming controls leads to interactivity and layout issues


Hazelmeow:
**When the offset of flet.Container and the scale property of its content are set, some buttons in the content are not clickable.**

**Code example to reproduce the issue:**

```python
import flet

def main(page: flet.Page):
    page.add(
        flet.Container(
            width = 300, height = 300,
            alignment = flet.alignment.center,
            content = flet.Column(
                alignment = flet.MainAxisAlignment.CENTER,
                controls = [
                    flet.ElevatedButton(text="Elevated button 1"),
                    flet.ElevatedButton(text="Elevated button 2"),
                    flet.ElevatedButton(text="Elevated button 3"),
                    flet.ElevatedButton(text="Elevated button 4"),
                    flet.ElevatedButton(text="Elevated button 5"),
                    flet.ElevatedButton(text="Elevated button 6"),
                ]
            )
        )
    )

def main2(page: flet.Page):
    page.add(
        flet.Container(
            width = 300, height = 300,
            scale = 2,
            offset = flet.transform.Offset(0.5, 0.5),
            alignment = flet.alignment.center,
            content = flet.Column(
                alignment = flet.MainAxisAlignment.CENTER,
                controls = [
                    flet.ElevatedButton(text="Elevated button 1"),
                    flet.ElevatedButton(text="Elevated button 2"),
                    flet.ElevatedButton(text="Elevated button 3"),
                    flet.ElevatedButton(text="Elevated button 4"),
                    flet.ElevatedButton(text="Elevated button 5"),
                    flet.ElevatedButton(text="Elevated button 6")
                ]
            )
        )
    )

def main3(page: flet.Page):
    page.add(
        flet.Container(
            width = 300, height = 300,
            offset = flet.transform.Offset(0.5, 0.5),
            alignment = flet.alignment.center,
            content = flet.Column(
                scale = 2,
                alignment = flet.MainAxisAlignment.CENTER,
                controls = [
                    flet.ElevatedButton(text="Elevated button 1"),
                    flet.ElevatedButton(text="Elevated button 2"),
                    flet.ElevatedButton(text="Elevated button 3"),
                    flet.ElevatedButton(text="Elevated button 4"),
                    flet.ElevatedButton(text="Elevated button 5"),
                    flet.ElevatedButton(text="Elevated button 6")
                ]
            )
        )
    )

flet.app(target = main) # works
flet.app(target = main2) # works
[...]

PR #1094, opened 2023-02-26
Ensure that a Container can fully display its content

iron3oxide:
Adds the `dimensions` property to `ConstrainedControl`, which calculates width x height after the scale is applied (if there is one). On second thought, `visual_dimensions` might be a more accurate name, but this is a draft anyways since tests and docs are missing. See #1073 for why this might be useful.


iron3oxide:
Forgot to do sync my fork properly, that's why the discussion template commits are in there. Sorry!

PR #1141, opened 2023-03-06
add simple django like routing

saurabhwadekar:
This is the first contribution of my life.
If I have done any mistake, then definitely tell me.
Thank you

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/__init__.py
+++ b/sdk/python/packages/flet-core/src/flet_core/__init__.py
@@ -171,3 +171,4 @@ from flet_core.user_control import UserControl
 from flet_core.vertical_divider import VerticalDivider
 from flet_core.view import View
 from flet_core.window_drag_area import WindowDragArea
+from flet_core.flet_route import Routing,path
--- a/sdk/python/packages/flet-core/src/flet_core/flet_route.py
+++ b/sdk/python/packages/flet-core/src/flet_core/flet_route.py
@@ -1,0 +1,58 @@
+from flet_core import Page,View
+import repath
+
+def path(url: str, clear: bool, view: View):
+    return [url, clear, view]
+
+
+
+class Routing:
+    """
+import flet as ft
+from flet_core import Routing,path
+from settings.routing import Routing
+
+def index_view(page: ft.Page,params={}):
+    "/",
+    return ft.view(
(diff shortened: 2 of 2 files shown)


ndonkoHenri:
Can you explain your PR? 
What is the use case? How is it useful etc.

saurabhwadekar:
# This makes dynamic routing easier.

Excuse me I don't know English. This has been translated by Google Translate.
Let me know if you don't understand anything

We have to write a lot of (if, else) statements for routing and passing the params to the view is very difficult.
But now all we need to do is to create a routing object in the main function and pass the page and routing list.
in routing list We'll use the path function. it returns a list
In this function we have to pass the routing url, in clear we have to (true/false) and in view we have to pass a function which will return ft.view


example:

## main.py
```python
import flet as ft
from flet_core import Routing,path
from views.index_view import IndexView
from views.next_view import NextView

def main(page: ft.Page):

    app_routes = [
        path(url="/", clear=True, view=IndexView),
        path(url="/next_view/:my_id", clear=False, view=NextView),
    ]

    Routing(page=page,app_routes=app_routes)
    page.go(page.route)

ft.app(target=main)

```

## views/index_view.py
```python
import flet as ft

def IndexView(page: ft.Page,params={}):
    print(params)
    return ft.view(
        "/",
        controls=[
            ft.Text("This Is Index View"),
            ft.ElevatedButton("Go Next View", on_click=lambda _: page.go("/next_view/10")),
        ]
    )

```

## views/next_view.py
```python
import flet as ft

def NextView(page: ft.Page,params={}):
    print(params)
    return ft.view(
        "/next_view/:my_id",
        controls=[
            ft.Text("This Is Next View"),
            ft.ElevatedButton("Go Index View", on_click=lambda _: page.go("/")),
        ]
    )

```

FeodorFitsner:
Thanks for your efforts making the PR!

As far as I see you don't need access to any Flet internals to make your functionality work. I think instead of adding that into Flet's core you could just release it as a Python module, right?

saurabhwadekar:
Yes it can be done. and I will definitely do it

berecik:
@saurabhwadekar FYI with the https://github.com/Marysia-Software-Limited/flet-django package you can use directly a Django routing in your flet project, even without a Django project

PR #1283, opened 2023-04-13
padding attribute is not zero by default from the view component

AdamMusa:
I noticed a small error in view component,the default padding attribute of view is None but when I launch my application I have the impression that padding is not none so I played with the padding attribute and I understood that padding takes a value of 10, I don't know where the value comes from so I decided to correct this little error by passing a default value to padding to zero.
Here is some screenshoot with not passing padding a value .
screen
view
view_package

After passing padding to zero as default value here is the screenshot of it 

final


FeodorFitsner:
Thanks for your efforts!

Unfortunately, we can't merge this PR because of [this](https://github.com/flet-dev/flet/pull/580#issuecomment-1317531012).

Similar to #904, #883

PR #1302, opened 2023-04-17
Capturing web client info: IP and user agent

### LINKED ISSUE #1053: Feature Request: More session info
**Please Describe The Problem To Be Solved**

When logging or rate limiting, or simply debugging which user is doing what, we need more information about the session:  IP address, browser info etc

**(Optional): Suggest A Solution**

I believe Flutter already provides this information.  I could be added to a Page variable, put into the page session data or better:

Flask provides an `from flask import g` and `g` is a dict that we can put anything in we want and its available anywhere by simply importing, you don't have to pass a `page` around to get it.  This would be the optimal solution to have session global data rather than using page.  in Flask g is guaranteed to be session specific.

FeodorFitsner:
New `page` properties: `client_ip` and `client_user_agent`. Close #1053

Usage example:

```python
import flet as ft

def main(page: ft.Page):

    page.add(
        ft.Text(f"Client IP: {page.client_ip}"),
        ft.Text(f"Client user agent: {page.client_user_agent}"),
    )

ft.app(target=main, port=8550, view=ft.WEB_BROWSER)
```

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/page.py
+++ b/sdk/python/packages/flet-core/src/flet_core/page.py
@@ -229,6 +229,8 @@ class Page(Control):
             Command(0, "get", ["page", "windowHeight"]),
             Command(0, "get", ["page", "windowTop"]),
             Command(0, "get", ["page", "windowLeft"]),
+            Command(0, "get", ["page", "clientIP"]),
+            Command(0, "get", ["page", "clientUserAgent"]),
         ]
 
     def __set_page_details(self, values):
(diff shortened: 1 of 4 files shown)


tokyohandsome:
How can we get those properties from a web app (`flet build web`)?

ndonkoHenri:
It is because you the result of Flet build web is a static website. 
In this state you can read neither the ip nor the client user agent.
Your app needs to be deployed as a web app for it to work but not as a static app.

PR #1382, opened 2023-05-09
update user api typing

Vlod-github:
`MainAxisAlignment` already had a default value, now it's some specific value. Same with `ImageFit`. Write your opinion. Depending on that, we'll decide what to do with the rest


FeodorFitsner:
I would move from `MainAxisAlignmentDefault` to specific values like `MainAxisAlignment.START`.
Because if it works with `MainAxisAlignment` it won't necessary work with e.g. `CrossAxisAlignment` because it defaults to `START` in one controls and `CENTER` to another. And you don't need to use default in `get_enum_from_value` as it's already assigned to a property in object constructor, right?
Could you fix python tests too please, to make CI build work?

FeodorFitsner:
Could you make tests passing please: https://ci.appveyor.com/project/flet-dev/flet/builds/47037311/job/aji7oskx3kkanq7j

Vlod-github:
Don't worry, I'll fix the tests. The past days have been loaded. Should finish this week

Vlod-github:
I understand what you wrote, now I will check the default value for each component in the documentation

Vlod-github:
@FeodorFitsner, see if you have any comments or suggestions, and then I can continue to adjust the other components

Vlod-github:
`container` `clip_behavior` - double default value. Perhaps it should be done differently

Vlod-github:
@FeodorFitsner 
To be honest, I do not know what the error is in the tests. Now I've done a rebase, but it didn't help. The last successful commit was `fix Image Repeat tests`. Perhaps this is a CI CD error

sdk/python/packages/flet-core/src/flet_core/alert_dialog.py
```diff
@@ -207,14 +210,8 @@ def actions_alignment(self) -> MainAxisAlignment:
...
-        self.__actions_alignment = value
-        if isinstance(value, MainAxisAlignment):
-            self._set_attr("actionsAlignment", value.value)
-        else:
-            self.__set_actions_alignment(value)
-
-    def __set_actions_alignment(self, value: MainAxisAlignmentString):
-        self._set_attr("actionsAlignment", value)
+        self.__actions_alignment = get_valid_enum(MainAxisAlignment, value, _MainAxisAlignmentDefault)
+        self._set_attr("actionsAlignment", self.__actions_alignment.value)
```
FeodorFitsner, line 214: @Vlod-github I'm thinking on the entire approach here...
At the moment the idea behind those `None` or `MainAxisAlignment.NONE` or etc. is that if the property is not set it's not being sent over the wire to the server (Fletd), i.e. something like `{"t": "alertdialog"}` is sent if no properties are set.

When we set something non-None in the constructor it goes to `self._set_attr()` and will be set if just `AlertDialog()` constructor is used, something like `{"t": "alertdialog", "actionsAlignment": "end"}`, thus increasing overall traffic by sending unnecessary data.

What if further improve the logic around enum props and will be calling `self._set_attr()` with None if semantically default enum value is specified, kind of:

```python
self._set_attr("actionsAlignment", self.__actions_alignment.value if self.__actions_alignment != _MainAxisAlignmentDefault else None)
```

Do you see the point? Does it make sense?

FeodorFitsner:
I fixed CI: https://ci.appveyor.com/project/flet-dev/flet - update from `main`.

Vlod-github:
@FeodorFitsner I've been thinking about it too. I think we need to answer the question "Can we guarantee that components will have the attributes declared in the documentation on any platform with any webview if we don't explicitly set default attributes on the backend?" Because most enterprise developers would rather have predictable behavior on all platforms instead of some network costs. Can we guarantee stability on the frontend side?

Vlod-github:
Very good. Different browsers can set different initial values of css properties for html elements. And it is important that in these different browsers or different browser engines components behave the same. If a component changes its visual behavior depending on the screen size, it is normal

Vlod-github:
Moved enumerations to flet_core/types and ran into cyclic import and other problems. While I'm waiting for the pylance developers to respond and then I'll finish it, there's not much left
https://github.com/microsoft/pylance-release/issues/4455

PR #1389, opened 2023-05-11
Add CLI option -m to run as module

### LINKED ISSUE #702: Flet CLI to run an app from a module?
I was wondering if its possible use the flet-cli with a module instead of a script? Thanks!

michasng:
This change resolves #702.
Adding `-m` applies the corresponding [python cli option](https://docs.python.org/3/using/cmdline.html#cmdoption-m).
In this case, the `script` is no longer a file path, but a module path, so `foo.fee` as opposed to `foo/fee.py`, hence in order to find the script_path, we need to convert the module path to a file path.
Note that this can also run directories as modules by running the `__main__.py` file inside of it.

### DIFF
--- a/sdk/python/packages/flet/src/flet/cli/commands/run.py
+++ b/sdk/python/packages/flet/src/flet/cli/commands/run.py
@@ -32,6 +32,14 @@ class Command(BaseCommand):
             help="custom TCP port to run Flet app on",
         )
         parser.add_argument(
+            "-m",
+            "--module",
+            dest="module",
+            action="store_true",
+            default=False,
+            help="treat the script as a python module path as opposed to a file path",
+        )
+        parser.add_argument(
             "-d",
             "--directory",
             dest="directory",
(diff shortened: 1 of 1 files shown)


sdk/python/packages/flet/src/flet/cli/commands/run.py
```diff
@@ -74,9 +82,17 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
 
     def handle(self, options: argparse.Namespace) -> None:
         # print("RUN COMMAND", options)
-        script_path = options.script
-        if not os.path.isabs(options.script):
-            script_path = str(Path(os.getcwd()).joinpath(options.script).resolve())
+        if options.module:
+            script_path = str(options.script).replace(".", "/")
+            if os.path.isdir(script_path):
+                script_path += "/__main__.py"
```
FeodorFitsner, line 88: Would `/` work on Windows?

michasng: Thankfully yes, I'm on windows and python can handle it apparently. Even when I mix `/` and `\` it seems.

SKbarbon: `os.path.join` ?

PR #1584, opened 2023-07-11
Added support for AES-256 bit GCM encryption

CPullr:
Despite the fact that AES-128 provides a solid amount of security I believe that going the extra step and adding support for the AES-256-GCM encryption and authentication scheme can be extremely useful to users who want more security in their flet project while not having to actually mess with outside cryptographic libraries. It uses the same libraries as the already existing cryptographic implementation (hashlib + cryptography) so it will not mess with dependencies.


CPullr:
It appears that I did not setup my Git Email before opening this pr so it picked up my computer's username as the Github user :sob: . Should I close and reopen this PR?

FeodorFitsner:
Yes, please! :)

PR #1585, opened 2023-07-11
Added support for AES-256 bit GCM encryption 

acidicly:
Despite the fact that AES-128 provides a solid amount of security I believe that going the extra step and adding support for the AES-256-GCM encryption and authentication scheme can be extremely useful to users who want more security in their flet project while not having to actually mess with outside cryptographic libraries. It uses the same libraries as the already existing cryptographic implementation (hashlib + cryptography) so it will not mess with dependencies.


CPullr:
Alr It accepted the signature!

PR #1608, opened 2023-07-19
fix changelog for the new import module for OAuth providers

04nada:
current changelog incorrect points to the new import for OAuth providers as being in `flet.auth.providers`, but it should be in `flet_runtime.auth.oauth_provider`

also, I feel that further emphasis for the "breaking change" could/should be made in the changelog for `flet_runtime` replacing `flet` in general for a number of modules, beyond just the OAuth providers


peterwwillis:
fyi @FeodorFitsner there are still pages on flet's website that reference the old (broken) method:
 - https://flet.dev/docs/guides/python/authentication/
 - https://flet.dev/blog/user-authentication

Took me a while of debugging/investigating to figure out the correct form

PR #1756, opened 2023-08-22
Hosting Flet web apps in FastAPI

FeodorFitsner:


ndonkoHenri:
Adding docstrings to
classes/methods/functions will be awesome... don't you think so?

FeodorFitsner:
You are absolutely right!

PR #1997, opened 2023-10-27
Put it the right version of flet 

AdamMusa:
I followed the contribution guide correctly and I tried to run a small application by running the command **poetry install** then **poetry run python hello.py** I came across this error. we assume I have a file with this content

```
import flet
from flet import Page, Text

def main(page: Page):
    page.add(Text("Hello, world!"))

flet.app(target=main)
```
after this  command  **poetry run python hello.py** here is the error 

error

I found the error is on this line of file flet/src/version.py
version = "0.2.0" and this version does not exist yet in https://github.com/flet-dev/flet/releases/
so I hard coded  with one exist version if all the condition not respecte and everything work correctly

result


FeodorFitsner:
Did you install Flet with `pip` or just cloned the repo?
Did you clone repo with `git clone` or downloaded as zip? Do you have a localized version of Git?

AdamMusa:
> Did you install Flet with `pip` or just cloned the repo? Did you clone repo with `git clone` or downloaded as zip? Do you have a localized version of Git?

I cloned the repo with git clone .

FeodorFitsner:
OK, what output do you get by running `git status` while in repo's directory?

AdamMusa:
> OK, what output do you get by running `git status` while in repo's directory?

On both branches(my branch and main branch) I got the same output 

**On the hand branch
Your branch is up to date with 'origin/main' or  'origin/fix_version'.

nothing to validate, the working copy is clean**

FeodorFitsner:
I guess this line is the culprit: https://github.com/flet-dev/flet/blob/238a37e0c0c63712d6f4c9b6fdca246200e9f130/sdk/python/packages/flet/src/flet/version.py#L23

AdamMusa:
> I guess this line is the culprit:

Yeah

FeodorFitsner:
But why "On the hand branch" in your case? Could you post a screenshot please? What version of git is that?

AdamMusa:
> But why "On the hand branch" in your case? Could you post a screenshot please? What version of git is that?

Oh  it was my fault it would be in "On main branch"

My system is in french 
Here is the screenshots of my branch and main branch
main
fix

AdamMusa:
I have not provided more necessary information?

FeodorFitsner:
I fixed it here, differently: https://github.com/flet-dev/flet/commit/091c9151e153ff6be651fec3774dd164791cbc17

AdamMusa:
I noticed you add code only for windows I hope it going to work on Mac . Because the bug I occured and fixed about Mac

FeodorFitsner:
It's a fix for all platforms. You can give a try with preview version of Flet.

PR #1865, opened 2023-09-23
Supporting `WebView` control

SKbarbon:
This pull request includes updates for both the Flet Flutter package and the Flet Python SDK. These updates introduce support for a WebView control on mobile platforms (iOS and Android) through the class named `MobileWebViewer`.
## peek
https://github.com/flet-dev/flet/assets/86029286/b5f4e062-0c1a-4581-be65-3d957b60af0d

## code example
```python
import flet

def main (page:flet.Page):
    page.appbar = flet.AppBar(
        title=flet.Text("WebView in flet!")
    )

    wv = flet.MobileWebViewer("http://localhost:3000/", width=400, height=650)
    page.add(wv)

flet.app(target=main)
```

## desktop suggestion
I propose we distinguish between WebView components for desktop and mobile platforms. Typically, desktop WebView contains tools tailored for desktop-class functionality, which may not be accessible on mobile due to its limited capabilities. Merging these two types of WebView may constrain the capabilities of the desktop WebView widget. To address this, I recommend intentionally separating them by naming the desktop WebView as `DesktopWebViewer` and keeping the mobile version as `MobileWebViewer`. This deliberate separation will help maintain the distinct features and capabilities of each platform.

PR #2049, opened 2023-11-08
`WebView` control

FeodorFitsner:
Original work by @SKbarbon in https://github.com/flet-dev/flet/pull/1865


rboyd:
we have an immediate use for this. thank you @SKbarbon and @FeodorFitsner !

FeodorFitsner:
@rboyd good! beware it's for iOS and Android only.

Issue #631, opened 2022-11-25
Dynamic Validations in `TextField`


ipritom:
How can I add validations in a `TextField`?

My Use cases:
`TextField` for username which won't allow space in a text.

`TextField` for number input which won't allow non-numeric input.

ndonkoHenri:
https://api.flutter.dev/flutter/material/TextFormField-class.html ?

FeodorFitsner:
There is no built-in validation mechanism into TextField control, but you can implement that in your program using `TextField.on_change` or `TextField.on_blur` events and set `TextField.error_text` into non-empty string to make it red.

ndonkoHenri:
Onchange is called after the keyboard value has been inserted into the field. 

It will look ugly to let the user see his text be inputted, then see it removed.

Think of it...we need in-built validators or the ability to overwrite the method of TextField inserting the the text. 

See kivy example: https://kivy.org/doc/stable/api-kivy.uix.textinput.html#filtering

JWDT:
@ndonkoHenri - I've tried a quick implementation of a filter by using `on_change`, and it appears to be *almost* fast enough (at least in a test case) that it doesn't render the character to the screen before removing it.

The problem I had though was changing the `control.value` then doing `page.update()` would move the cursor back to the start of the text input, so for example, if "a" was being filtered out, typing the word "bad" would result in "db" instead of "bd".

Validation in the same way web browsers tend to do it, i.e. allowing you to type anything you want into the box, then checking it and simply complaining (as @FeodorFitsner is suggesting), and perhaps blocking any type of "submit" action looks to be a workable solution that users might be more familiar with.

I know I would be more frustrated with an app for ignoring my input -- or worse, changing something I pasted in -- than it telling me the input isn't valid and letting me fix it (or even having a button to filter it for me *after* I've typed what I want).

ndonkoHenri:
That's right.
That approach sounds better to me now, after reading your comment.

FeodorFitsner:
> I know I would be more frustrated with an app for ignoring my input -- or worse, changing something I pasted in...

Yeah, you are trying to type something into the field and it doesn't respond and then after a few minutes you like "oh, it's a numeric field!". Agree, showing validation errors on blur or submit is more user friendly.

wpritom:
> @ndonkoHenri - I've tried a quick implementation of a filter by using `on_change`, and it appears to be _almost_ fast enough (at least in a test case) that it doesn't render the character to the screen before removing it.


[...]

PR #2101, opened 2023-11-18
Feature: `TextField` Input validation

ndonkoHenri:
Close https://github.com/flet-dev/flet/issues/631

**Props:**
```python
@dataclasses.dataclass
class InputFilter:
    regex_string: str
    allow: bool = field(default=True)
    replacement_string: str = field(default="")
```

**Example code:**
```python
import flet as ft


def main(page: ft.Page):
    page.title = "Input Filtering"
    page.add(
        ft.TextField(
            label="Only numbers are allowed :)",
            input_filter=ft.InputFilter(allow=True, regex_string=r"[0-9]"), # only allow numbers
        )
    )

ft.app(target=main)

```

**Capture:**
(when trying to add text characters they are rejected)
example

**Notes:**
- `regex_string` can't be passed to dart as a raw string(using the 'r') 
  ```dart
  RegExp(regexString ?? "")
  ```
  As seen above, this is because the value of `regex_string` is stored in a variable (`regexString`).. hence using string interpolation(`RegExp(r"$regexString")`) for example, will cant work because the variable's name will not be interpolated (we are in raw mode). 
From [`RegExp`](https://api.flutter.dev/flutter/dart-core/RegExp-class.html) docs:
> Use a raw string to treat each character, including \ and $, in a string as a literal character. Each character then gets passed to the [RegExp](https://api.flutter.dev/flutter/dart-core/RegExp-class.html) parser. You should use a raw string as the argument to the [RegExp](https://api.flutter.dev/flutter/dart-core/RegExp-class.html) constructor.

This is my first dart contribution, hence will strongly appreciate opinions. 😄

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/textfield.py
+++ b/sdk/python/packages/flet-core/src/flet_core/textfield.py
@@ -71,6 +71,14 @@ class InputFilter:
     allow: bool = field(default=True)
     replacement_string: str = field(default="")
 
+class NumbersOnlyInputFilter(InputFilter):
+    def __init__(self):
+        super().__init__(r"[0-9]")
+
+class TextOnlyInputFilter(InputFilter):
+    def __init__(self):
+        super().__init__(r"[a-zA-Z]")
+
 
 class TextField(FormFieldControl):
     """
--- a/sdk/python/packages/flet-core/src/flet_core/textfield.py
+++ b/sdk/python/packages/flet-core/src/flet_core/textfield.py
@@ -1,3 +1,5 @@
+import dataclasses
+from dataclasses import field
 import time
 from enum import Enum
 from typing import Any, Optional, Union
@@ -61,6 +63,13 @@ class TextCapitalization(Enum):
     CHARACTERS = "characters"
     WORDS = "words"
     SENTENCES = "sentences"
+
+
+@dataclasses.dataclass
+class InputFilter:
+    allow: bool = field(default=True)
+    regex_string: str = field(default="[0-9]")
+    replacement_string: str = field(default="")
 
 
 class TextField(FormFieldControl):
@@ -179,6 +188,7 @@ class TextField(FormFieldControl):
         cursor_height: OptionalNumber = None,
         cursor_radius: OptionalNumber = None,
         selection_color: Optional[str] = None,
+        input_filter: Optional[InputFilter] = None,
         on_change=None,
         on_submit=None,
         on_focus=None,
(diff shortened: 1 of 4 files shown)


sdk/python/packages/flet-core/src/flet_core/textfield.py
```diff
@@ -63,6 +65,13 @@ class TextCapitalization(Enum):
     SENTENCES = "sentences"
 
 
+@dataclasses.dataclass
+class InputFilter:
+    allow: bool = field(default=True)
+    regex_string: str = field(default="[0-9]")
```
FeodorFitsner, line 71: I would remove default from `regex_string` field and require developer to explicitly specify the pattern. Defaults for `allow=True` and `replacement_string=""` look good.

ndonkoHenri: Alright, diving into that.

ndonkoHenri: What do you think of this?

```python
regex_string: str = field(default="")
```

sdk/python/packages/flet-core/src/flet_core/textfield.py
```diff
@@ -63,6 +65,13 @@ class TextCapitalization(Enum):
     SENTENCES = "sentences"
 
 
+@dataclasses.dataclass
+class InputFilter:
+    regex_string: str
```
FeodorFitsner, line 70: Looks good! If you want to "help" people you can add a few inherited from `InputFilter` classes with pre-defined regex, e.g. `NumbersOnlyInputFilter`, etc.

FeodorFitsner:
Well, looks like it's ready to merge, huh?

sdk/python/packages/flet-core/src/flet_core/textfield.py
ndonkoHenri: Nice idea!
Will work on that later today.

ndonkoHenri:
Yh, now I think its ready!

bobwatcherx:
i get eror for use inputfilter

```
Unhandled error processing page session : Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/flet_runtime/app.py", line 363, in on_session_created
    session_handler(page)
  File "/home/minx/belajar/fletval/main.py", line 9, in main
    input_filter=ft.InputFilter(allow=True, regex_string=r"[0-9]", replacement_string=""),
AttributeError: module 'flet' has no attribute 'InputFilter'
```
flet version

```
root@minx-Lenovo-V145-14AST:/home/minx/belajar/fletval# flet --version
0.12.2
```
code

```
import flet as ft


def main(page: ft.Page):
    page.title = "Input Filtering"
    page.add(
        ft.TextField(
    label="Only numbers are allowed :)",
    input_filter=ft.InputFilter(allow=True, regex_string=r"[0-9]", replacement_string=""),
) 
   )

ft.app(target=main)
```

ndonkoHenri:
You can get it from the prereleases. An official release is still to come.

```bash
pip install flet --pre
```

Issue #2100, opened 2023-11-18
Page::Session::Set Doesn't work when using hot reloading.


xzripper:
**Description**

When running application with hot loader, many strange issues happening: `Page::Session::Set` doesn't work (simply doesn't  registers key), issues with `Page::Go`. When running application normally it works very good.

**Code example to reproduce the issue:**

```python
def register(error_message):
    username_str = username.value
    email_str = email.value
    password_str = password.value

    if len(username_str) <= 5:
        error_message.value = 'Username should be at least 5 characters.'

        pg.page.update()

        return

    if not match(r'^[_a-z0-9-]+(\.[_a-z0-9-]+)*@[a-z0-9-]+(\.[a-z0-9-]+)*(\.[a-z]{2,4})$', email_str):
        error_message.value = 'Invalid Email.'

        pg.page.update()

        return

    if len(password_str) <= 8:
        error_message.value = 'Password should be at least 8 characters.'

        pg.page.update()

        return

    username_busy = False
    email_busy = False

    for user in db.get_value('users'):
        if user.username == username_str: username_busy = True
        if user.email == email_str: email_busy = True

    if username_busy and email_busy:
        error_message.value = 'Username and Email is already busy.'

        pg.page.update()

        return

    elif username_busy:
        error_message.value = 'Username is already busy.'

        pg.page.update()

        return

    elif email_busy:
        error_message.value = 'Email is already busy.'

        pg.page.update()

        return

    user = UserModel(username_str, email_str, password_str, [])

    db.get_value('users').append(user)

    db.commit()

    pg.page.session.set('user', username_str) # Here issue happens.

    pg.navigator.navigate('/', pg.page) # And here.
```

**Describe the results you received:**
`user` key wasn't registered, and page hasn't changed.

**Describe the results you expected:**
I expected it to register `user` key and change page to `/`.

**Additional information you deem important (e.g. issue happens only occasionally):**
This issues happening only in hot reload mode.

**Flet version (`pip show flet`)**:

```
0.10.3
```

**Operating system**:
Windows 10

xzripper:
Hot reload:
![chrome_bXd84gZVss](https://github.com/flet-dev/flet/assets/94743980/692e66ae-3786-4b24-bab9-e1fe2dcaa180)

No hot reload:
![chrome_9WEfT08NFD](https://github.com/flet-dev/flet/assets/94743980/548a0a81-7e35-45c8-bb57-cf12ab26f772)

[...]

PR #2105, opened 2023-11-18
Ignore directories while using hot reloading.

xzripper:
```
set IGNORE_DIRS=dir\to\ignore;dir2\to\ignore

/ or /

$env:IGNORE_DIRS=dir\to\ignore;dir2\to\ignore

/ or /

export IGNORE_DIRS=dir\to\ignore;dir2\to\ignore
```

If a file from these directories is updated, no event is fired and the application is not restarted.

#2100 - Issue.

### DIFF
--- a/sdk/python/packages/flet/src/flet/cli/commands/run.py
+++ b/sdk/python/packages/flet/src/flet/cli/commands/run.py
@@ -258,14 +258,18 @@ class Handler(FileSystemEventHandler):
         th.start()
 
     def on_any_event(self, event):
-        if (
-            self.script_path is None or event.src_path == self.script_path
-        ) and not event.is_directory:
-            current_time = time.time()
-            if (current_time - self.last_time) > 0.5 and self.is_running:
-                self.last_time = current_time
-                th = threading.Thread(target=self.restart_program, args=(), daemon=True)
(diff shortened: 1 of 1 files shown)


FeodorFitsner:
Thanks for the PR!

How about doing that with another argument in `flet run` command, for example `--ignore`?

FeodorFitsner:
A few notes:

1. It's more optimal to initialize `ignore_files` in the constructor of the handler class rather than in `on_any_event()`.
2. `--ignore` option files and directories must be resolved relative to `script_dir`, not a current directory.

I can run with `flet run /path/to/my/app --ignore assets;data` and I'd like `/path/to/my/app/assets` and `/path/to/my/app/data` ignored.

3. When working with file paths use `pathlib`, `os.path` or other path-manipulation libraries instead of working with strings and slashes. It should be cross-platform, but `\` is Windows only.
4. I'd use [os.path.commonpath](https://stackoverflow.com/a/37095733/1435891) method to check if `event.src_path` is a sub-directory of a path in `ignore_files` collection. This way `ignore_files` is just absolute paths of items passed in `--ignore` option.

FeodorFitsner (approved):

PR #2231, opened 2023-12-12
CupertinoSwitch control

InesaFitsner:


package/lib/src/controls/switch.dart
```diff
@@ -1,3 +1,5 @@
+import 'package:flet/src/controls/cupertino_switch.dart';
```
FeodorFitsner, line 1: Replace with relative path to a file.

FeodorFitsner (approved):

PR #2244, opened 2023-12-18
Fix FLET_VIEW_PATH ignored on linux

mse11:


FeodorFitsner:
Why is it ignored on Linux?

FeodorFitsner:
I see, thanks for the PR!

FeodorFitsner (approved):

PR #2259, opened 2023-12-22
Refactor code, add Enum deprecation utils

ndonkoHenri:
`ThemeVisualDensity.ADAPTIVEPLATFORMDENSITY` is now deprecated, and `ThemeVisualDensity.ADAPTIVE_PLATFORM_DENSITY` should be used instead.
The deprecation class is in the `theme.py` file.

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/utils.py
+++ b/sdk/python/packages/flet-core/src/flet_core/utils.py
@@ -6,8 +6,6 @@ import re
 import string
 import sys
 import unicodedata
-from enum import EnumMeta
-from warnings import warn
 
 
 def random_string(length):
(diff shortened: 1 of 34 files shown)


sdk/python/packages/flet-core/src/flet_core/utils.py
```diff
@@ -159,3 +161,15 @@ def with_degrees(self, value):
         radians = math.radians(value)
         magnitude = abs(self)
         return type(self).polar(radians, magnitude)
+
+
+class DeprecatedEnumMeta(EnumMeta):
+
```
FeodorFitsner, line 167: I would put that class next to `ThemeVisualDensity` (in the same `[src/flet_core/theme.py](https://github.com/flet-dev/flet/pull/2259/files#diff-24599f89970071ace465e406200accb4cdd5a8be6c55fe1a1f1e25618db7516c)` file) and call it `ThemeVisualDensityDeprecated`. What do you think?

ndonkoHenri: Leaving as-is will permit us to use this class on any Enum (add new ifs/cases to the getattribute method), just by passing it as metadata(see VisualDensity). Wdyt?

FeodorFitsner: My point is `DeprecatedEnumMeta` is not targeting specific enum, i.e. you are checking just for `item == 'ADAPTIVEPLATFORMDENSITY'` without checking what enum is that. Imagine the switch will grow up for more obsolete enum members.

I suggest having "Deprecated" sibling for every enum that has deprecated members and keep that class next to its enum.

ndonkoHenri: Ah, i get your point now! :)
Will give it a look.

FeodorFitsner (approved):

PR #2270, opened 2023-12-27
Added TextStyle for the Text control

50Bytes-dev:
Added TextStyle support for the Text control and added a height attribute for TextStyle

### DIFF
--- a/package/lib/src/utils/text.dart
+++ b/package/lib/src/utils/text.dart
@@ -173,11 +173,6 @@ TextStyle textStyleFromJson(ThemeData theme, Map<String, dynamic> json) {
     decorations.add(TextDecoration.lineThrough);
   }
 
-  var height = json["height"];
-  if (height != null && !(height is double)) {
-    height = double.tryParse(height.toString());
-  }
-
   return TextStyle(
       fontSize: json["size"] != null ? parseDouble(json["size"]) : null,
       fontWeight: fontWeight != null ? getFontWeight(fontWeight) : null,
@@ -186,7 +181,7 @@ TextStyle textStyleFromJson(ThemeData theme, Map<String, dynamic> json) {
           : null,
       fontFamily: json["font_family"],
       fontVariations: variations,
-      height: height,
+      height: json["height"] != null ? parseDouble(json["height"]) : null,
       decoration:
           decorations.isNotEmpty ? TextDecoration.combine(decorations) : null,
       decorationStyle: json["decoration_style"] != null
--- a/package/lib/src/utils/text.dart
+++ b/package/lib/src/utils/text.dart
@@ -173,6 +173,11 @@ TextStyle textStyleFromJson(ThemeData theme, Map<String, dynamic> json) {
     decorations.add(TextDecoration.lineThrough);
   }
 
+  var height = json["height"];
+  if (height != null && !(height is double)) {
+    height = double.tryParse(height.toString());
+  }
+
   return TextStyle(
       fontSize: json["size"] != null ? parseDouble(json["size"]) : null,
       fontWeight: fontWeight != null ? getFontWeight(fontWeight) : null,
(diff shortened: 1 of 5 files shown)


sdk/python/packages/flet-core/src/flet_core/control.py
```diff
@@ -481,6 +482,9 @@ def _build_command(self, update=False):
                 sval = str(val).lower()
             elif isinstance(val, dt.datetime) or isinstance(val, dt.date):
                 sval = val.isoformat()
+            elif dataclasses.is_dataclass(val):
+                sval = dataclasses.asdict(val)
+                sval = json.dumps(sval)
             else:
```
FeodorFitsner, line 488: I like the idea of serializing dataclass attributes centrally - thanks for the idea!

But at the moment we are doing that in `_before_build_command()` method, like for example here: https://github.com/flet-dev/flet/blob/main/sdk/python/packages/flet-core/src/flet_core/text_span.py#L45-L47

I'd stick to this method for now, until we decide to migrate to your method - I don't know how it's going to affect existing functionality 🧐

sdk/python/packages/flet-core/src/flet_core/text.py
```diff
@@ -137,7 +138,7 @@ def __init__(
         size: OptionalNumber = None,
         weight: Optional[FontWeight] = None,
         italic: Optional[bool] = None,
-        style: Optional[TextThemeStyle] = None,
+        style: Optional[Union[TextThemeStyle, TextStyle]] = None,
```
FeodorFitsner, line 141: Nice idea! I'm not a big fan of mixing types though, but let's users decide! :)

sdk/python/packages/flet-core/src/flet_core/control.py
50Bytes-dev: @FeodorFitsner Thank you for your feedback. It's nice to be helpful for such a cool project. Have you considered pydantic as a serializer? I could help with rewriting the code using pydantic.

FeodorFitsner: Absolutely, I'd love to look into improving the thing with Pydantic - let's kick off the discussion after New Year ;)
For your PR we could still apply old-school approach :)

sdk/python/packages/flet-core/src/flet_core/text.py
ndonkoHenri: I think we might need to separate props... perhaps create `text_style` and `theme_style`?

FeodorFitsner: Since @ndonkoHenri noticed this 🫣 I'm not going to try convincing him that "it's OK like that" - let's have two separate properties 😏

Now, I'd love to have `style` property be of `TextStyle` to look the same as `TextSpan.style`, and then `theme_style` for named styles.

However, we don't want to break a lot of existing applications. I'm thinking what if we "temporary" make `style` property accepting both `TextThemeStyle` and `TextStyle` (like in this PR) and also add `theme_style` which accepts only `TextThemeStyle`. Will update docs for `.style` property to allow only `TextStyle`. What do you guys think?

FeodorFitsner: @50Bytes-dev would you implement that (along with previous note) please?

50Bytes-dev:
@FeodorFitsner check it

sdk/python/packages/flet-core/src/flet_core/control.py
FeodorFitsner: Could you change to `_before_build_command()` please?

50Bytes-dev: > Could you change to `_before_build_command()` please?

complete

package/lib/src/utils/text.dart
```diff
@@ -173,6 +173,11 @@ TextStyle textStyleFromJson(ThemeData theme, Map<String, dynamic> json) {
     decorations.add(TextDecoration.lineThrough);
   }
 
+  var height = json["height"];
+  if (height != null && !(height is double)) {
+    height = double.tryParse(height.toString());
+  }
+
```
FeodorFitsner, line 180: You can use the same approach as for `json["size"]` (see below), can't you?

50Bytes-dev: > You can use the same approach as for `json["size"]` (see below), can't you?

complete

FeodorFitsner (approved):

PR #2286, opened 2023-12-30
`Text.style` type Deprecation warning

### LINKED ISSUE #2284: Add deprecation warning in `Text.style` setter

ndonkoHenri:
Close #2284 

The warning message is shown only when setting the style as below (after instance creation):
```py
import flet as ft
t = ft.Text(
    "Test!",
    # style=ft.TextThemeStyle.HEADLINE_SMALL,
)
t.style = ft.TextThemeStyle.HEADLINE_SMALL

```
When setting it at the Text instance creation, the warning is weirdly not shown.

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/text.py
+++ b/sdk/python/packages/flet-core/src/flet_core/text.py
@@ -274,7 +274,6 @@ class Text(ConstrainedControl):
             warn(
                 "If you wish to set the TextThemeStyle, use `Text.theme_style` instead. "
                 "The `Text.style` property should be used to set the TextStyle only.",
-                DeprecationWarning,
                 stacklevel=2,
             )
 
--- a/sdk/python/packages/flet-core/src/flet_core/text.py
+++ b/sdk/python/packages/flet-core/src/flet_core/text.py
@@ -1,6 +1,7 @@
 import dataclasses
 from enum import Enum
 from typing import Any, List, Optional, Union
+from warnings import warn
 
 from flet_core.constrained_control import ConstrainedControl
 from flet_core.control import OptionalNumber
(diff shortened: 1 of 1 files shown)


FeodorFitsner (approved):

FeodorFitsner:
Ha ha, I've asked ChatGPT about that and got this:

>However, certain types of warnings, such as DeprecationWarning, are ignored by default.

>If you’re not seeing the warnings, it’s possible that the warning is being suppressed. This can happen if the warning filter, which is a sequence of matching rules and actions, is set to ignore the type of warning you’re trying to issue.

By adding a filter I can now see DeprecationWarnings:

```python
import warnings

import flet as ft

warnings.simplefilter("always", DeprecationWarning)

t = ft.Text(
    "Test!",
    style=ft.TextThemeStyle.HEADLINE_SMALL,
)
```

I'm not sure I like this Python default bahavior. I don't think someone will guess to enable warnings.

If `DeprecationWarning` category is removed from `warn()` method the warning is shown as `UserWarning` in all cases. Shall we remove it then?

ndonkoHenri:
> If DeprecationWarning category is removed from warn() method the warning is shown as UserWarning in all cases. Shall we remove it then?

I tried your suggestion, and two warnings (instead of just one) are now shown:
```
/Users/ndonkohenri/PycharmProjects/flet-dev/flet/sdk/python/packages/flet-core/src/flet_core/text.py:172: UserWarning: If you wish to set the TextThemeStyle, use `Text.theme_style` instead. The `Text.style` property should be used to set the TextStyle only.
  self.style = style
/Users/ndonkohenri/PycharmProjects/flet-dev/flet/sdk/python/playground/test-changes.py:8: UserWarning: If you wish to set the TextThemeStyle, use `Text.theme_style` instead. The `Text.style` property should be used to set the TextStyle only.
  t.style = ft.TextThemeStyle.HEADLINE_SMALL
```
This is not a problem though. Should I nevertheless move on commiting the removed DeprecationWarning.

FeodorFitsner:
Two warnings is correct behaviour as the 1st call is in constructor and 2nd in property setter.
Yes, let's remove `DeprecationWarning`.

FeodorFitsner (approved):

PR #2337, opened 2024-01-08
add template route with defaults

johnnychen94:
Found the `TemplateRoute` util when I made https://github.com/flet-dev/website/pull/199

I add this because I feel it's more intuitive to write

```python
troute = TemplateRoute(page.route, defaults={"id": get_random_id()})
if troute.match("/books/:id?"):
    # if the route is "/books", the `id` property will be set to the random ID
    print("Book view ID:", troute.id)
else:
    print("Unknown route")
```

than using `hasattr`:

```python
troute = TemplateRoute(page.route)
if troute.match("/books/:id?"):
    # if the route is "/books", the `id` property will be set to the random ID
    id = troute.id if hasattr(troute, "id") else "1"
    print("Book view ID:", troute.id)
else:
    print("Unknown route")
```

Maybe the `defaults` should be added to the `troute.match("/books/:id", id="1")` instead? I'm not sure.

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/template_route.py
+++ b/sdk/python/packages/flet-core/src/flet_core/template_route.py
@@ -3,9 +3,44 @@ import repath
 
 
 class TemplateRoute:
-    def __init__(self, route: str) -> None:
+    """
+    A route helper class that allows you to match a route template against a route string.
+
+    Example:
+
+    ```python
+    troute = TemplateRoute(page.route)
+
+    if troute.match("/books/:id"):
+        print("Book view ID:", troute.id)
+    elif troute.match("/account/:account_id/orders/:order_id"):
+        print("Account:", troute.account_id, "Order:", troute.order_id)
+    else:
+        print("Unknown route")
+    ```
+
+    When the `:id` part is optional, you can use the `?` suffix the , e.g.,
+
+    ```python
+    troute = TemplateRoute(page.route, defaults={"id": get_random_id()})
+    if troute.match("/books/:id?"):
+        # if the route is "/books", the `id` property will be set to the random ID
+        print("Book view ID:", troute.id)
+    else:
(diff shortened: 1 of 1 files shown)


sdk/python/packages/flet-core/src/flet_core/template_route.py
```diff
@@ -18,7 +53,15 @@ def match(self, route_template: str) -> bool:
 
         if match:
             self.__last_params = match.groupdict()
-            for k, v in self.__last_params.items():
-                setattr(self, k, v)
+            # optional pattern such as `:id?` will be matched and set as None
+            # if there are defaults provided, we skip overwriting the properties with None
+            skip_none = True if self.__defaults else False
+            self.__set_props(self.__last_params, skip_none=skip_none)
```
johnnychen94, line 59: Second thought. This is, strictly speaking not an incremental change because the behavior of the following code changes:

```python
troute = TemplateRoute(page.route, defaults={"id": get_random_id()})
if troute.match("/books/:name?"):
    print("Book view ID:", troute.name)
else:
    print("Unknown route")
```

In the main branch `troute.name is None` and in this branch `troute.name` does not exist.

I'd like to know if you guys are in favor of this "defaults" idea (whether it's added to the `__init__` or to the `match`), and then I'll update accordingly.

FeodorFitsner: You mean your addition breaks some existing scenarios?

Also, why would you need to set book ID to a random number? 😏 Could you provide a real-world scenario?

johnnychen94: > Could you provide a real-world scenario?

Good question! I admit that I don't have any based on my limited app dev experiences.

I think I'll just close this PR.

PR #2362, opened 2024-01-10
Fix types

taaaf11:
Fix types in code blocks

@ndonkoHenri Please review this.


ndonkoHenri:
Somehow related to https://github.com/flet-dev/website/pull/129.

Just for info: `TextField.value` can accept integer or string. Nevertheless, I think for consistency throughout (docs, type definition) we should use strings. (as in this PR)

taaaf11:
Yes, I was thinking about this too, that the person who created [flet-dev/website#129](https://github.com/flet-dev/website/pull/129) may think that I just saw their code. But it's not related to that PR. I just made the types consistency

PR #2364, opened 2024-01-10
Fix types and make code consistent with docs.

taaaf11:
In the docs [flet.dev/docs](https://flet.dev/docs/), `view=ft.AppView.WEB_BROWSER` is used. But here in the README.md, `flet.WEB_BROWSER` is used. I know it works, but it's not consistent with the docs.


taaaf11:
@FeodorFitsner Kindly review my code

FeodorFitsner (approved):

PR #2368, opened 2024-01-10
Fixed: PubSub is not shared between pages in the same FastAPI app

### LINKED ISSUE #2367: PubSub is not shared between pages in the same FastAPI app
### Discussed in https://github.com/flet-dev/flet/discussions/2361

<sup>Originally posted by **rogeremail** January  9, 2024</sup>
### Question

After I modified the example code to be asynchronous and then used the flet_fastapi, I only received messages from the sender.

### Code sample

```python
import flet as ft
import flet_fastapi

async def main(page: ft.Page):
    page.title = "Flet Chat"

    async def on_message_async(msg):
        messages.controls.append(ft.Text(msg))
        await page.update_async()

    await page.pubsub.subscribe_async(on_message_async)

    async def send_click(e):
        await page.pubsub.send_all_async(f"{message.value}")
        message.value = ""
        await page.update_async()

    messages = ft.Column()
    message = ft.TextField(hint_text="Your message...", expand=True)
    send = ft.ElevatedButton("Send", on_click=send_click)
    await page.add_async(messages, ft.Row(controls=[message, send]))

#ft.app(target=main, view=ft.AppView.WEB_BROWSER) # This will work, but I need to deploy to the server 
app = flet_fastapi.app(main) # After I modified the example code to be asynchronous and then used the flet_fastapi, I only received messages from the sender.
```

### Error message

_No response_

### ------------------------------------------------------

- [X] I have searched for answers to my question both in the issues and in previous discussions.</div>

FeodorFitsner:
Fix #2367

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/page.py
+++ b/sdk/python/packages/flet-core/src/flet_core/page.py
@@ -390,6 +390,7 @@ class Page(Control):
                 await c.will_unmount_async()
 
     def _close(self):
+        self.__pubsub.unsubscribe_all()
         removed_controls = self._remove_control_recursively(self.index, self)
         for c in removed_controls:
             c.will_unmount()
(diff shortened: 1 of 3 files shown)


ndonkoHenri (approved):
Tried, and it now works as expected.

Issue #2473, opened 2024-01-24
pack command should not delete "dist" folder if specified another with "--distpath"


IagoBeuller:
I specify the a folder named "pack" to save the executable with --distpath, and instead of deleting "pack" folder, it is deleting "dist", where was deleted lots of important files. Luckly The folder is from google drive, and I could recover the files.

taaaf11:
We are working on the issue.

taaaf11:
@FeodorFitsner I have created #2500 pull request to solve the issue... Kindly review the pr

taaaf11:
@IagoBeuller The issue is now resolved. Kindly close the issue. Install pre-release (obviously in a virtual environment) or wait for the next release ;)

PR #2500, opened 2024-01-28
Bugfix: `flet pack --distpath` deletes `dist` directory

taaaf11:
`flet pack --distpath DISTPATH` deletes "dist" directory even when `DISTPATH` is specified. The solution is to ask the user whether to delete the directories.

Related issue #2473 .

### DIFF
--- a/sdk/python/packages/flet/src/flet/cli/commands/pack.py
+++ b/sdk/python/packages/flet/src/flet/cli/commands/pack.py
@@ -116,15 +116,36 @@ class Command(BaseCommand):
         )
 
     def handle(self, options: argparse.Namespace) -> None:
+        is_dir_not_empty = lambda dir: os.path.isdir(dir) and len(os.listdir(dir)) != 0
+
         # delete "build" directory
         build_dir = os.path.join(os.getcwd(), "build")
-        if os.path.exists(build_dir):
-            shutil.rmtree(build_dir, ignore_errors=True)
+        if is_dir_not_empty(build_dir):
+            ask_to_delete = input('Do you want to delete "build" directory? (y/n) ')
+            if not ask_to_delete.lower() == "n":
+                shutil.rmtree(build_dir, ignore_errors=True)
+            else:
+                print('Failing... "build" directory must be empty to proceed.')
+                exit(1)
 
-        # delete "dist" directory
-        dist_dir = os.path.join(os.getcwd(), "dist")
-        if os.path.exists(dist_dir):
-            shutil.rmtree(dist_dir, ignore_errors=True)
+        # delete "dist" directory or DISTPATH directory
+        # if --distpath cli option is specified
+        if options.distpath:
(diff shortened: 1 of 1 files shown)


taaaf11:
@FeodorFitsner Kindly review my pull request

sdk/python/packages/flet/src/flet/cli/commands/pack.py
```diff
@@ -116,15 +116,36 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
...
 
     def handle(self, options: argparse.Namespace) -> None:
+        is_dir_not_empty = lambda dir: os.path.isdir(dir) and len(os.listdir(dir)) != 0
+
         # delete "build" directory
         build_dir = os.path.join(os.getcwd(), "build")
-        if os.path.exists(build_dir):
-            shutil.rmtree(build_dir, ignore_errors=True)
+        if is_dir_not_empty(build_dir):
+            ask_to_delete = input('Do you want to delete "build" directory? (y/n) ')
```
FeodorFitsner, line 124: Does `build` directory have any value to care about its deletion? A developer won't be doing/customizing there anything?

sdk/python/packages/flet/src/flet/cli/commands/pack.py
```diff
@@ -116,15 +116,36 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
...
 
-        # delete "dist" directory
-        dist_dir = os.path.join(os.getcwd(), "dist")
-        if os.path.exists(dist_dir):
-            shutil.rmtree(dist_dir, ignore_errors=True)
+        if is_dir_not_empty(dist_dir):
+            ask_to_delete = input(
+                f'Do you want to delete "{os.path.basename(dist_dir)}" directory? (y/n) '
+            )
+            if not ask_to_delete.lower() == "n":
```
FeodorFitsner, line 142: When we introduce interactive prompts to a CLI I believe we should add some sort of `--force` or `--yes` option, so the question is skipped in non-interactive mode, right?

sdk/python/packages/flet/src/flet/cli/commands/pack.py
taaaf11: Just in case, the user has their some important files in `build` directory. I would remove this if you want it to be removed by the way,,

sdk/python/packages/flet/src/flet/cli/commands/pack.py
taaaf11: > I believe we should add some sort of --force or --yes

Working on the issue 👍

taaaf11: Made the necessary changes @FeodorFitsner 👍

FeodorFitsner (approved):

PR #2574, opened 2024-02-08
added letter_spacing attribute for TextStyle

ariefendi992:
Example code
```py

import flet
from flet import Page, Text, TextStyle


def main(page: Page):
    page.window_always_on_top = True
    page.add(
        Text(
            "Hello, world!\nHello, Flet",
            style=TextStyle(letter_spacing=6),
        )
    )
    page.update()


flet.app(target=main)

```

### DIFF
--- a/package/lib/src/controls/text.dart
+++ b/package/lib/src/controls/text.dart
@@ -78,23 +78,26 @@ class TextControl extends StatelessWidget with FletStoreMixin {
       }
 
       style = (style ?? const TextStyle()).copyWith(
-          fontSize: control.attrDouble("size", null),
-          fontWeight: getFontWeight(fontWeight),
-          fontStyle: control.attrBool(
-            "italic",
-            false,
-          )!
-              ? FontStyle.italic
-              : null,
-          fontFamily: control.attrString("fontFamily"),
-          fontVariations: variations,
-          color: HexColor.fromString(
-                  Theme.of(context), control.attrString("color", "")!) ??
-              (spans.isNotEmpty
-                  ? DefaultTextStyle.of(context).style.color
-                  : null),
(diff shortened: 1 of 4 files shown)


ndonkoHenri (changes requested):
Thanks for your contribution!
The two properties you added are TextStyle properties - so I suggest we add them to the [`TextStyle` dataclass](https://github.com/ariefendi992/flet/blob/d45ca11c99ff05c9bee70d34ece68f21074e7547/sdk/python/packages/flet-core/src/flet_core/text_style.py#L28) instead.

FeodorFitsner (approved):

package/lib/src/controls/text.dart
```diff
@@ -78,23 +78,25 @@ class TextControl extends StatelessWidget with FletStoreMixin {
...
+        fontFamily: control.attrString("fontFamily"),
+        fontVariations: variations,
+        color: HexColor.fromString(
+                Theme.of(context), control.attrString("color", "")!) ??
+            (spans.isNotEmpty
+                ? DefaultTextStyle.of(context).style.color
+                : null),
+        backgroundColor: HexColor.fromString(
+            Theme.of(context), control.attrString("bgcolor", "")!),
+        letterSpacing: control.attrDouble('letterSpacing', null),
```
ndonkoHenri, line 98: Having it here is unnecessary because `Text` has no more has `letter_spacing` attribute.

ndonkoHenri (approved):

ndonkoHenri:
Thanks, @ariefendi992 ! I am merging...
Can you please make a PR [here](https://github.com/flet-dev/website), adding this new property to the `TextStyle` docs?

ariefendi992:
yes, i can.

Issue #2628, opened 2024-02-14
Container Issue when ```ink = True```


lekshmanmj:
**Description**
Border, bgcolor, shadow, ink animation are hidden below other controls when inked container is placed among other controls in a stack.

**Code example to reproduce the issue:**

```python
import flet as ft

def main(page: ft.Page):
    page.vertical_alignment = ft.MainAxisAlignment.CENTER
    page.horizontal_alignment = ft.CrossAxisAlignment.CENTER
    page.theme_mode = ft.ThemeMode.DARK
    button = ft.Container(
        bgcolor= ft.colors.GREEN,
        content=ft.Icon(
            ft.icons.MENU_BOOK_ROUNDED,
            color= ft.colors.BLUE,
            size= 30
        ),
        padding=18,
        alignment=ft.alignment.center,
        border_radius=20,
        border= ft.Border(
            top= ft.BorderSide(width= 2, color= ft.colors.RED),
            left= ft.BorderSide(width= 2, color= ft.colors.RED),
            bottom= ft.BorderSide(width= 2, color= ft.colors.RED),
            right= ft.BorderSide(width= 2, color= ft.colors.RED)
        ),
        shadow = ft.BoxShadow(
            blur_radius = 20,
            color = ft.colors.YELLOW_ACCENT_700,
            blur_style = ft.ShadowBlurStyle.OUTER,
        ),
        on_click=lambda e: print("Clicked"),
        ink=True,
        bottom= 20, right= 20
    )
    simply_container = ft.Container(
        bgcolor= ft.colors.with_opacity(1, ft.colors.BLACK),
        height= page.height * 0.8,
        width= page.width * 0.8,
        bottom= 20, right= 40
    )
    stack = ft.Stack(
        controls= [
            simply_container,
            button
        ],
        expand= True
    )
    page.add(stack)

ft.app(target=main)

```

**Describe the results you received:**
![image](https://github.com/flet-dev/flet/assets/35882740/9e7ad25e-617a-4bc4-9b48-1adc911a03e7)

**Describe the results you expected:**
![image](https://github.com/flet-dev/flet/assets/35882740/61dca2ac-d751-45a0-abf0-cd84f42deab1)

**Additional information you deem important (e.g. issue happens only occasionally):**
This Issue occurs only when ```ink = True```

**Flet version (`pip show flet`)**:

```
Name: flet
Version: 0.19.0
Summary: Flet for Python - easily build interactive multi-platform apps in Python
Home-page: 
[...]

PR #2684, opened 2024-02-20
Add more `InkWell` event in container.dart

isaffathir:
### InkWell Control from Flutter
_https://api.flutter.dev/flutter/material/InkWell-class.html_

- [x] Revert change to default (Duplicate InkWell control)
- [x] Add on_release event (Callback after click)
- [x] Add ink_color event (Able to change ink color)

**Test Code:**
```python
import flet as ft

def main(page: ft.Page):
    page.add(
        ft.Container(
            content=ft.Text(
                "Login",
                style=ft.TextStyle(
                    size=14,
                    color=ft.colors.GREY_700,
                    weight=ft.FontWeight.BOLD,
                ),
            ),
            margin=ft.margin.only(left=30, right=30, top=10),
            padding=20,
            alignment=ft.alignment.center,
            bgcolor=ft.colors.GREY_100,
            border_radius=10,
            ink=True,
            ink_color=ft.colors.CYAN,
            on_release=lambda e: print("onrelease"),
            on_click=lambda e: print("onclick"),
        ),
    )

ft.app(target=main)
```


FeodorFitsner:
It does look like a duplicate of `Container` control with `ink=True`.

lekshmanmj:
@isaffathir You may be interest in solving this issue, Please have a look at this issue https://github.com/flet-dev/flet/issues/2628

isaffathir:
> It does look like a duplicate of `Container` control with `ink=True`.

i look into Container and yes it use `ink = True` for InkWell, but what if i want to make ink for Text and Icon,or any?
thats InkWell control works, it will make any widget clickable. Its seems like `gesture_detector.py` do,but gesturedetector doesnt have splash/ripple effect when we `on_tap` the widget

**Make Text clickable by `InkWell` :**
```python
ft.InkWell(
                    on_tap=lambda e: print("Sign Up"),
                    content=ft.Text("Click me for Sign Up", color=ft.colors.BLUE),
                ),
```

**Make Text clickable by `GestureDetector` :**
ft.GestureDetector(
                    on_tap=lambda e: print("Sign Up"),
                    content=ft.Text("Click me for Sign Up", color=ft.colors.BLUE),
                ),

FeodorFitsner:
You can put any control inside Container and make it clickable, no?

isaffathir:
> You can put any control inside Container and make it clickable, no?

It does,but i just suggest to make other more flexible to create clickable widget, just like `gesture_detector.py` do. or should i add the other `InkWell` event on `container.dart`?

FeodorFitsner:
I mean `InkWell` is already part of Container when `ink=True`: https://github.com/flet-dev/flet/blob/main/packages/flet/lib/src/controls/container.dart#L153

and you can wrap container with gesture detector as well.

isaffathir:
> I mean `InkWell` is already part of Container when `ink=True`: https://github.com/flet-dev/flet/blob/main/packages/flet/lib/src/controls/container.dart#L153
> 
> and you can wrap container with gesture detector as well.

I got it, should i add other `InkWell` event such a `splashColor`,`highlightColor`,etc on `container.dart`? but i think if python InkWell exist is more better right?

ndonkoHenri:
Hey @isaffathir! Great work on this PR.
The [Flet-idea](https://flet.dev/docs/#what-is-flet) is actually not to implement the Flutter API one-to-one:
> Flet does not just "wrap" Flutter widgets, but adds its own "opinion" by combining smaller widgets, hiding complexities, implementing UI best-practices, applying reasonable defaults

I read [this](https://stackoverflow.com/a/56725359/21791365) answer from stack-overflow, and it seems like `GestureDetector` and `Inkwell` are exactly thesame, with main difference that `Inkwell` offers a customisable ripple effect, while GD doesn't.
However, you may not have noticed it, but the Flet `Container` control, already wraps/makes use of `Ink` and `Inkwell` internally as seen below:
https://github.com/flet-dev/flet/blob/fe1a4d15fb31065231941ad737da551f8c6e9940/packages/flet/lib/src/controls/container.dart#L148-L153

So, to be brief, the only thing Flet lacks atm is the possibility of customising the ripple effect. The solution will then be to further expose `Ink`/`Inkwell` properties (related to ripple) found in this `container.dart`.  

If you find the addition of this interesting/useful, you could perhaps work in that direction instead. 🙂

isaffathir:
Thanks for answer @ndonkoHenri, your answer make me know Flet further. in this case i should add other `InkWell `event such a `splashColor`,`highlightColor`,etc on `container.dart`?

ndonkoHenri:
> in this case i should add other `InkWell `event such a `splashColor`,`highlightColor`,etc on `container.dart`

Definitely! Go on...

FeodorFitsner:
Close as dup of https://github.com/flet-dev/flet/pull/2701

PR #2701, opened 2024-02-21
Fix container.dart for issue #2628

isaffathir:
**Fix for Issue :** #2628 
The `Ink` or `InkWell` widget must have a [Material](https://api.flutter.dev/flutter/material/Material-class.html) widget as an ancestor.

seems `container.dart` doesn't have:
https://github.com/flet-dev/flet/blob/fe1a4d15fb31065231941ad737da551f8c6e9940/packages/flet/lib/src/controls/container.dart#L151-L153

so i try to wrap `InkWell` with `Material` :
https://github.com/flet-dev/flet/pull/2701/files#diff-3d2babe2e48d8eaf9b1c35294b8c498c905a03e16f8e490258b215a98e5ec568R151-R154

i also remove `Ink` cause Flet already use `Container` as parent, because `Container` can take a `BoxDecoration` as i read [this](https://stackoverflow.com/a/52228086)

This is how i fix this issue, hopefully it will help.

### DIFF
--- a/packages/flet/lib/src/controls/container.dart
+++ b/packages/flet/lib/src/controls/container.dart
@@ -66,6 +66,7 @@ class ContainerControl extends StatelessWidget with FletStoreMixin {
         children.where((c) => c.name == "content" && c.isVisible);
     bool ink = control.attrBool("ink", false)!;
     bool onClick = control.attrBool("onclick", false)!;
+    bool onTap = control.attrBool("ontap", false)!;
     String url = control.attrString("url", "")!;
     String? urlTarget = control.attrString("urlTarget");
     bool onLongPress = control.attrBool("onLongPress", false)!;
@@ -145,7 +146,7 @@ class ContainerControl extends StatelessWidget with FletStoreMixin {
 
       Widget? result;
 
-      if ((onClick || url != "" || onLongPress || onHover) &&
+      if ((onTap || onClick || url != "" || onLongPress || onHover) &&
           ink &&
           !disabled) {
         var ink = Material(
@@ -155,7 +156,12 @@ class ContainerControl extends StatelessWidget with FletStoreMixin {
               // Dummy callback to enable widget
               // see https://github.com/flutter/flutter/issues/50116#issuecomment-582047374
               // and https://github.com/flutter/flutter/blob/eed80afe2c641fb14b82a22279d2d78c19661787/packages/flutter/lib/src/material/ink_well.dart#L1125-L1129
-              onTap: onHover ? () {} : null,
+              onTap: onTap
+                  ? () {
+                      debugPrint("Container ${control.id} Tap!");
+                      backend.triggerControlEvent(control.id, "tap", "");
+                    }
+                  : null,
               onTapDown: onClick || url != ""
                   ? (details) {
                       debugPrint("Container ${control.id} clicked!");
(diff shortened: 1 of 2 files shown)


isaffathir:
https://github.com/flet-dev/flet/assets/88388332/18bee2ca-5b41-4c8b-b75e-2229f2511995

**Test Code:**
```python
import flet as ft

def main(page: ft.Page):
    page.vertical_alignment = ft.MainAxisAlignment.CENTER
    page.horizontal_alignment = ft.CrossAxisAlignment.CENTER
    page.theme_mode = ft.ThemeMode.DARK
    button = ft.Container(
        bgcolor=ft.colors.GREEN,
        content=ft.Icon(ft.icons.MENU_BOOK_ROUNDED, color=ft.colors.BLUE, size=30),
        padding=18,
        alignment=ft.alignment.center,
        border_radius=20,
        border=ft.Border(
            top=ft.BorderSide(width=2, color=ft.colors.RED),
            left=ft.BorderSide(width=2, color=ft.colors.RED),
            bottom=ft.BorderSide(width=2, color=ft.colors.RED),
            right=ft.BorderSide(width=2, color=ft.colors.RED),
        ),
        shadow=ft.BoxShadow(
            blur_radius=20,
            color=ft.colors.YELLOW_ACCENT_700,
            blur_style=ft.ShadowBlurStyle.OUTER,
        ),
        on_click=lambda e: print("Clicked"),
        ink=True,
        bottom=20,
        right=20,
    )
    simply_container = ft.Container(
        bgcolor=ft.colors.with_opacity(1, ft.colors.BLACK),
        height=page.height * 0.8,
        width=page.width * 0.8,
        bottom=20,
        right=40,
    )

    stack = ft.Stack(controls=[simply_container, button], expand=True)
    page.add(stack)


ft.app(target=main)

```

ndonkoHenri:
Thanks @isaffathir. 

But each new/non-existing dart property you add should be "exposed" in Flet's python API. Just like you did in the first PR.

isaffathir:
> Thanks @isaffathir.
> 
> But each new/non-existing dart property you add should be "exposed" in Flet's python API. Just like you did in the first PR.

there is nothing new, i just wrap `InkWell` with `Material` from `material.dart` that already exist. let me know if im wrong answer or misunderstand

isaffathir:
> Thanks @isaffathir.
> 
> But each new/non-existing dart property you add should be "exposed" in Flet's python API. Just like you did in the first PR.

`Material` is imported from `material.dart`
https://github.com/flet-dev/flet/blob/fe1a4d15fb31065231941ad737da551f8c6e9940/packages/flet/lib/src/controls/container.dart#L5

i just wrap `InkWell` with `Material`. is this new? let me know if im wrong answer or misunderstand

lekshmanmj:
Thanks @isaffathir for fixing the bug and that too immediately when I mentioned you with this issue.

packages/flet/lib/src/controls/container.dart
```diff
@@ -148,8 +148,9 @@ class ContainerControl extends StatelessWidget with FletStoreMixin {
       if ((onClick || url != "" || onLongPress || onHover) &&
           ink &&
           !disabled) {
-        var ink = Ink(
-            decoration: boxDecor,
+        var ink = Material(
+            color: Colors.transparent,
```
ndonkoHenri, line 152: Why transparent?

ndonkoHenri:
> i just wrap InkWell with Material. are this new? let me know if im wrong answer or misunderstand

Ah, i forgot to check which issue you were trying to fix.

isaffathir:
As suggestion from [this](https://api.flutter.dev/flutter/material/Ink-class.html) that need transparent Material also [here](https://stackoverflow.com/a/76688164), yes we can use `Ink` if don't want use `Material` instead,but because flet add `Container` as parent for the return using `Ink` as a child will cause #2628 problem,thats why need transparent `Material ` as ancestor

PR #2906, opened 2024-03-18
`Control.parent` property

### LINKED ISSUE #952: `Control.parent` property
### Discussed in https://github.com/flet-dev/flet/discussions/950

<sup>Originally posted by **ndonkoHenri** January 29, 2023</sup>
@FeodorFitsner, what do you think of adding a way to access the **parent** of a control?
I thought of this idea because alot of guys have been asking how to delete an item from a `ListView` (`ListTile`) when a button/Control in the tile is pressed.
  The way I solve this issue was to always add a reference to the tile in the `data` property of the `Control` which will be used to delete the tile... and when the required event happens, grab the `e.control.data`(the ref), and `remove` it from the `ListView.controls` list. Nevertheless, I think a better, efficient, and alternative approach could be to access the parent of that Control `e.control.parent`(the ref), and `remove` it from the `ListView.controls` list.

**Why could this be useful?**
Using references could become an issue or fail when the `Control` keeping that reference is moved elsewhere. Whereas, using the `.parent` is cleaner and will never fail, because when the `Control` we need is moved any where, it's `.parent` is changed to its new parent.

I maybe poorly explained my thoughts, but `.parent` could be used in many other areas.

Let me know what you think of this. 
I just had the idea, but don't even know if it will be possible to implement :)

</div>

ndonkoHenri:
Closes #952

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/control.py
+++ b/sdk/python/packages/flet-core/src/flet_core/control.py
@@ -46,6 +46,7 @@ class Control:
         self.__data: Any = None
         self.data = data
         self.__event_handlers = {}
+        self.parent: Control = None
         if ref:
             ref.current = self
 
@@ -515,6 +516,8 @@ class Control:
                 indent=indent + 2, index=index, added_controls=added_controls
             )
             commands.extend(childCmd)
+            # set the parent
+            control.parent = self
 
         self.__previous_children.clear()
         self.__previous_children.extend(children)


sdk/python/packages/flet-core/src/flet_core/control.py
```diff
@@ -515,6 +516,8 @@ def _build_add_commands(self, indent=0, index=None, added_controls=None):
                 indent=indent + 2, index=index, added_controls=added_controls
             )
             commands.extend(childCmd)
+            # set the parent
+            control.parent = self
```
FeodorFitsner, line 520: That function where it's set is recursive, so `control.parent` will be set to a top most control which `update()` is called for.

ndonkoHenri: What is your suggestion?

ndonkoHenri:
> Issue TBF: The tiles created on_click of the first tile have no parent (`None`) although they all live in thesame `Column`.

Resolved! Property works like a charm with the updated test code. :)

FeodorFitsner (approved):

PR #2934, opened 2024-03-23
add global context for session

50Bytes-dev:
I was inspired by streamlit library, it has session_state, which is very convenient for me. I made something similar.

```python
# main.py
import flet as ft
from .utils import test_global_context, test_global_context_async

async def main(page: ft.Page):
    print(ft.context.page.session_id)
    test_global_context()
    page.run_thread(test_global_context)
    page.run_task(test_global_context_async)

# utils.py
import flet as ft

async def test_global_context_async():
    print("Work async!", ft.context.page.session_id)

def test_global_context():
    print("Work!", ft.context.page.session_id)
```


FeodorFitsner:
Thanks for your PR!

I've been wondering if your solution is the same as https://flet.dev/docs/guides/python/session-storage/ ?

50Bytes-dev:
> Thanks for your PR!
> 
> I've been wondering if your solution is the same as https://flet.dev/docs/guides/python/session-storage/ ?

No. My solution allows you to access the session without accessing the page variable. I solved the problem when I had to throw the page variable everywhere to access the session. Now you just need to ```import flet as ft``` and access ```ft.context.page``` from any file in the application.

50Bytes-dev:
For example, my application uses the langfuse library, which has a huge pile of tools and now instead of throwing a page variable into each tool I can just import flet and access flet.context.page from any tool.

FeodorFitsner:
OK, interesting. Never used context vars before - time to learn! :)

sdk/python/packages/flet-core/src/flet_core/page.py
```diff
@@ -51,12 +52,20 @@
...
+from flet_core.utils import deprecated, classproperty
 from flet_core.utils.concurrency_utils import is_pyodide
 from flet_core.view import View
 
 logger = logging.getLogger(flet_core.__name__)
 
+_session_page = ContextVar("flet_session_page", default=None)
+
+
+class context:
```
FeodorFitsner, line 64: Shouldn't be `context` exported in __init__ to be available as `ft.context`?

50Bytes-dev: I'm not sure I understand you correctly. As I find it very convenient to refer to context via `ft`. I use `ft` for all controls in my code.

FeodorFitsner: I mean `context` class should be imported in flet-core's `__init__.py`, no?

50Bytes-dev: Oh, that's right. My mistake. Forgot to add it to this branch)
Give me five minutes and I'll fix it.

50Bytes-dev: Done

FeodorFitsner (approved):

ndonkoHenri:
@50Bytes-dev, will you mind documenting this addition somewhere in the flet-docs? :)

shenzifanhy:
🆒

PR #2974, opened 2024-04-03
Add cookie for page variable

50Bytes-dev:
```python
import flet as ft

async def session_handler(page: ft.Page):
    if page.cookie:
        cookies = {name: value for name, value in [value.split("=") for value in page.cookie.split("; ")]}
        token = cookies.get("token")
        
ft.app(session_handler)
```

### DIFF
--- a/packages/flet/lib/src/flet_server.dart
+++ b/packages/flet/lib/src/flet_server.dart
@@ -49,6 +49,7 @@ class FletServer implements FletControlBackend {
   String _platform = "";
   String _platformBrightness = "";
   String _media = "";
+  String _cookie = "";
   int reconnectStarted = 0;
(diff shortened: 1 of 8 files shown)


ndonkoHenri:
> `{name: value for name, value in [value.split("=") for value in page.cookie.split("; ")]}`

To prevent the devs from having to do this, we could internally return either a `dict` with all the items or an empty dict `{}` when there's no cookie (or `None`?).

50Bytes-dev:
> I see it's for reading cookie only? It would be great to have a full-featured API for reading/writing cookies 😏

I did for my own needs. If I have time, I will expand the functionality

FeodorFitsner:
Oki-doki, I'm closing the PR then in favour of "future" full-scale implementation for cookies.

PR #3009, opened 2024-04-09
Allow arguments to be passed onto Stack

Sly0511:
Fixes the issue where passing arguments to UserControl no longer propagates into Stack subclassing


FeodorFitsner:
Do we need `*args` as well?

Sly0511:
Perhaps you are right, someone might be using non kwargs allow me a second

FeodorFitsner (approved):

Issue #3072, opened 2024-04-21
Video lag and stream interruptions on Android TV devices


bl1nch:
**Description**

High resolution videos play laggy on Android TV devices.
This problem is related to `media_kit` and can be found here: https://github.com/media-kit/media-kit/issues/614.
Possible solution: https://github.com/media-kit/media-kit/issues/614#issuecomment-2067968413.
If this solution actually works as I understand `flet_video` should be updated like this: https://github.com/flet-dev/flet/commit/240897e47b9d35bbee33f52fc9b3d297aef4d7e4

**Flet version (`pip show flet`)**:
```
0.22.0
```

**Operating system**:
```
Android TV
```

ndonkoHenri:
Can you open a PR with the commit?

PR #3074, opened 2024-04-21
Add `Video.configuration` property, instance of `VideoConfiguration` class

bl1nch:
https://github.com/media-kit/media-kit/issues/614
https://github.com/flet-dev/flet/issues/3072


ndonkoHenri:
Thanks!

I guess we have to merge this before you can try right?

FeodorFitsner:
You can try PR without merging:

image

ndonkoHenri:
I meant @bl1nch, not me. 😉

ndonkoHenri (approved):
Let's merge and see if it works...

bl1nch:
@ndonkoHenri sorry I'm new in this project, how I can build apk with these changes to test it?

ndonkoHenri:
Will let you know when you can test it.

FeodorFitsner:
Maybe we should implement `VideoControllerConfiguration` as a property in `Video` control? I read there could be issues when playing video in simulator. It would be great to control that depending on the platform. I'm afraid we could break something with this PR for other users.

ndonkoHenri:
> Maybe we should implement VideoControllerConfiguration as a property in Video control

Nice idea!
@bl1nch, do you want to further work on this?

bl1nch:
> > Maybe we should implement VideoControllerConfiguration as a property in Video control
> 
> Nice idea!

Sorry, have you already started implementing this? I haven't written in flutter or dart before, I'm afraid I won't be able to implement it in the near future.

ndonkoHenri:
> I'm afraid I won't be able to implement it in the near future.

No worries, thanks for your contribution! Will wrap this up :)

ndonkoHenri:
Created `VideoConfiguration` class, and added a new parameter to the Video control named `configuration`.
@FeodorFitsner, Thanks again for your suggestion! On macOS, when `output_backend="gpu"`,  a new window pops up with  video (window at the top in image below) 😬
Bildschirmfoto 2024-05-08 um 13 37 49

sdk/python/packages/flet-core/src/flet_core/video.py
```diff
@@ -178,6 +187,9 @@ def before_update(self):
         if isinstance(self.__subtitle_configuration, VideoSubtitleConfiguration):
             self._set_attr_json("subtitleConfiguration", self.__subtitle_configuration)
 
+        if isinstance(self.configuration, VideoConfiguration):
```
FeodorFitsner, line 190: I'd use `self.__configuration` here instead of `self.configuration` like everywhere else.

ndonkoHenri: Nice catch! 👍🏾

sdk/python/packages/flet-core/src/flet_core/video.py
```diff
@@ -178,6 +187,9 @@ def before_update(self):
         if isinstance(self.__subtitle_configuration, VideoSubtitleConfiguration):
             self._set_attr_json("subtitleConfiguration", self.__subtitle_configuration)
 
+        if isinstance(self.__configuration, VideoConfiguration):
+            self._set_attr_json("configuration", self.configuration)
```
FeodorFitsner, line 191: Could you fix that to `self.__configuration` as well. Sorry for the late reply.

ndonkoHenri: 🫣 - Thanks!

PR #3172, opened 2024-05-02
[reopen] Rename `NavigationDestination` to `NavigationBarDestination` + some type hinting, fixes and new features

bleudev:
Checklist:

- [x] Navigation bar's `NavigationDestination` renamed to `NavigationBarDestination`.
- [x] Type hinting:
    - [x] Add type alias:
        - [x] Number
        - [x] ControlEventFunction
        - [x] IconStr
        - [x] ColorStr
        - [x] Wrapper
    - [x] `navigation_bar.py`, `navigation_drawer.py`, `navigation_rail.py`
    - [x] `cupertino_navigation_bar.py`
    - [x] `page.py`
    - [x] `control.py`
- [x] Move `OptionalNumber` to `types.py`
- [x] `page.py`:
    - [x] Add `close_snack_bar() method`
    - [x] #3173
    - [x] Add Exceptions when page controls is None in `Page.close_...()` methods 
    - [x] Rename `show_...()` to `open_...()`
- [x] Magic methods:
    - [x] `Control in Page`

- [x] `isort` and `black` formatting

(closes #3173)
(closes #3329)
(reopen of #3161)

It should be merged with [#281](https://github.com/flet-dev/website/pull/281)

PR #3175, opened 2024-05-03
Error handling enhancements

ndonkoHenri:
- Py end: add assertion checks
- Dart end: make error messages clearer

### DIFF
--- a/packages/flet/lib/src/controls/cupertino_app_bar.dart
+++ b/packages/flet/lib/src/controls/cupertino_app_bar.dart
@@ -62,7 +62,7 @@ class CupertinoAppBarControl extends StatelessWidget
           : null,
       trailing: trailingCtrls.isNotEmpty
           ? createControl(control, trailingCtrls.first.id, control.isDisabled,
-              parentAdaptive: parentAdaptive)
+              parentAdaptive: parentAdaptive) // todo: wrap in row?
           : null,
       backgroundColor: bgcolor,
     );
--- a/packages/flet/lib/src/controls/dismissible.dart
+++ b/packages/flet/lib/src/controls/dismissible.dart
@@ -38,16 +38,19 @@ class _DismissibleControlState extends State<DismissibleControl> {
     bool disabled = widget.control.isDisabled || widget.parentDisabled;
     bool? adaptive =
         widget.control.attrBool("adaptive") ?? widget.parentAdaptive;
-    var contentCtrls = widget.children.where((c) => c.name == "content");
+    var contentCtrls =
+        widget.children.where((c) => c.name == "content" && c.isVisible);
 
     if (contentCtrls.isEmpty) {
-      return const ErrorControl("Dismissible does not have a content.");
+      return const ErrorControl(
+          "Dismissible.content must be provided and visible");
     }
 
-    var backgroundCtrls = widget.children.where((c) => c.name == "background");
+    var backgroundCtrls =
+        widget.children.where((c) => c.name == "background" && c.isVisible);
 
-    var secondaryBackgroundCtrls =
-        widget.children.where((c) => c.name == "secondaryBackground");
+    var secondaryBackgroundCtrls = widget.children
+        .where((c) => c.name == "secondaryBackground" && c.isVisible);
 
     var dismissThresholds =
         parseDismissThresholds(widget.control, "dismissThresholds");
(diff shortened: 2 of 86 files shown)


FeodorFitsner:
image 

🤯🤯🤯 Looking...

FeodorFitsner:
I honestly don't think those now-required properties are "breaking". Without them controls don't have any sense, right?

packages/flet/lib/src/controls/cupertino_app_bar.dart
```diff
@@ -62,7 +62,7 @@ class CupertinoAppBarControl extends StatelessWidget
           : null,
       trailing: trailingCtrls.isNotEmpty
           ? createControl(control, trailingCtrls.first.id, control.isDisabled,
-              parentAdaptive: parentAdaptive)
+              parentAdaptive: parentAdaptive) // todo: wrap in row?
```
FeodorFitsner, line 65: Why?

packages/flet/lib/src/controls/dismissible.dart
```diff
@@ -38,16 +38,19 @@ class _DismissibleControlState extends State<DismissibleControl> {
...
 
     if (contentCtrls.isEmpty) {
-      return const ErrorControl("Dismissible does not have a content.");
+      return const ErrorControl(
+          "Dismissible.content must be provided and visible");
     }
 
-    var backgroundCtrls = widget.children.where((c) => c.name == "background");
+    var backgroundCtrls =
+        widget.children.where((c) => c.name == "background" && c.isVisible);
```
FeodorFitsner, line 50: Thought maybe we need another helper for `.children.where((c) => c.name == "<name>" && c.isVisible)`?

packages/flet/lib/src/controls/cupertino_app_bar.dart
ndonkoHenri: In the conversion from `AppBar` to `CupertinoAppBar` when using `adaptive=True`, only the first trailing item is used/created - because the `CupertinoAppBar` allows only one child.
Was thinking of wrapping `AppBar.trailing` into a `Row`, then instead pass this row as child, making all the trailing controls visible.

packages/flet/lib/src/controls/dismissible.dart
ndonkoHenri: I don't seem to get it.
Helper in which sense?

ndonkoHenri:
> I honestly don't think those now-required properties are "breaking". Without them controls don't have any sense, right?

I considered as breaking in the sense that "any code without those properties will break".
Ex: I had to modify some of the test files because they were some tests were "failing" as a result of those required-parameters being missing.

sdk/python/packages/flet-core/src/flet_core/cupertino_sliding_segmented_button.py
```diff
@@ -107,14 +107,20 @@ def _before_build_command(self):
         super()._before_build_command()
         self._set_attr_json("padding", self.__padding)
 
+    def did_mount(self):
+        super().did_mount()
+        assert (
+            len(self.__controls) >= 2
+        ), "CupertinoSlidingSegmentedButton must have at minimum two visible controls"
+
```
FeodorFitsner, line 115: Why do you do checks in `did_mount()`? Would it be better do that in `control.before_update()` (new method), *before* updating UI? `did_mount` is when control has already been added.

sdk/python/packages/flet-core/src/flet_core/drag_target.py
```diff
@@ -149,11 +147,12 @@ def _get_control_name(self):
...
-        if self.__content:
-            self.__content._set_attr_internal("n", "content")
-            children.append(self.__content)
-        return children
+        self.__content._set_attr_internal("n", "content")
+        return [self.__content]
+
+    def did_mount(self):
+        super().did_mount()
+        assert self.__content.visible, "content must be visible"
```
FeodorFitsner, line 156: Yeah, `did_mount` is not the right place to do the checks. Let's fix it everywhere?

sdk/python/packages/flet-core/src/flet_core/range_slider.py
```diff
@@ -75,8 +75,8 @@ def main(page: ft.Page):
 
     def __init__(
         self,
-        start_value: [float],
-        end_value: [float],
+        start_value: float,
+        end_value: float,
```
FeodorFitsner, line 79: `Number` is coming in https://github.com/flet-dev/flet/pull/3172. We might use it everywhere once it's merged.

packages/flet/lib/src/controls/cupertino_app_bar.dart
FeodorFitsner: OK, let's keep it this way.

FeodorFitsner:
Great work in general! We need those checks and API polishing in general so much.
Only checks in `did_mount` should be changed.

PR #3186, opened 2024-05-03
Global Code Refactoring/Clean-up

ndonkoHenri:

PR #3263, opened 2024-05-10
Refactor `numbers.dart` utils

ndonkoHenri:
Part of #3186

## Tasks
The below utility methods are refactored to return `null` as default value. 
- [x] `parseInt`
- [x] `parseDouble`
- [x] `parseBool`

See `utils/themes.dart` for the main motivation behind this refactoring.


packages/flet/lib/src/utils/responsive.dart
```diff
@@ -21,7 +21,7 @@ Map<String, double> parseResponsiveNumber(
 }
 
 Map<String, double> responsiveNumberFromJson(Map<String, dynamic> json) {
-  return json.map((key, value) => MapEntry(key, parseDouble(value)));
+  return json.map((key, value) => MapEntry(key, parseDouble(value, 0)!));
```
FeodorFitsner, line 24: Why do we need explicit default values now? For clarity?

ndonkoHenri: Yep!
And also/mainly because the function now returns `null` as default, instead of `0`.

FeodorFitsner:
Could you check/resolve a conflict here please?

FeodorFitsner (approved):

PR #3298, opened 2024-05-15
Add `AutoComplete.selected_index` read-only property

ndonkoHenri:
## Test Code
```py
import flet as ft


def main(page: ft.Page):
    page.add(
        ft.AutoComplete(
            suggestions=[
                ft.AutoCompleteSuggestion(key="one 1", value="One"),
                ft.AutoCompleteSuggestion(key="two 2", value="Two"),
                ft.AutoCompleteSuggestion(key="three 3", value="Three"),
            ],
            on_select=lambda e: print(e.control.selected_index),
        )
    )


ft.app(target=main)


```


FeodorFitsner (approved):

fasilwdr:
Would it be possible to incorporate an additional property as a value in the event that a user inputs a value not selected from the provided suggestions?
@ndonkoHenri

ndonkoHenri:
I think it is possible, @fasilwdr.
Can you please open an issue(feature-request)?

PR #3406, opened 2024-05-31
Cleanup

ndonkoHenri:


sdk/python/packages/flet-core/src/flet_core/date_picker.py
```diff
@@ -161,12 +161,17 @@ def _get_control_name(self):
     def before_update(self):
         super().before_update()
 
+    @deprecated(
+        reason="Use Page.open() method instead.",
```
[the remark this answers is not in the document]
ndonkoHenri: @FeodorFitsner what do you think?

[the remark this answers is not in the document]
FeodorFitsner: how can we argue with AI? :) I dunno, let's leave it? We have the same situation in file picker too, right?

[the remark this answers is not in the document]
FeodorFitsner: Personally, I don't mind if the same result can be achieved with multiple methods.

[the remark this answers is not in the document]
ndonkoHenri: > how can we argue with AI?

Haha! The funniest thing with AI is, if you ask for a reason to remove this method, it is gonna give you one. 😂
I personnally think we should remove and keep only one way of opening.
> We have the same situation in file picker too, right?

Thats different actually. It has no open prop. The methods it has are useful and unique.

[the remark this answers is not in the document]
FeodorFitsner: OK, let's deprecate it then and leave `open` only.

sdk/python/packages/flet-core/src/flet_core/page.py
```diff
@@ -1000,311 +1009,251 @@ def __on_invoke_method_result(self, e) -> None:
...
+                    else:
+                        self.end_drawer = control
+                else:
+                    if self.drawer == control:
+                        control.update()
+                        return
+                    else:
+                        self.drawer = control
+                self.update()  # called only if the new drawer is different from the current one
+            elif control not in self.overlay:
```
FeodorFitsner, line 1031: For clarity:

```suggestion
            elif control not in self.__offstage.controls:
```

sdk/python/packages/flet-core/src/flet_core/page.py
```diff
@@ -1000,311 +1009,251 @@ def __on_invoke_method_result(self, e) -> None:
...
+                self.__offstage.update()
+            return
 
-        if control:
+    @staticmethod
+    def close(control: Control) -> None:
+        if hasattr(control, "open"):
             control.open = False
-            base.update()
+            control.update()
```
FeodorFitsner, line 1040: Shall we remove control from `self.__offstage.controls`?

sdk/python/packages/flet-core/src/flet_core/page.py
```diff
@@ -2210,6 +2181,8 @@ def banner(self) -> Optional[Banner]:
     @banner.setter
     def banner(self, value: Optional[Banner]):
         self.__banner = value
+        if value is not None:
+            self.__controls.append(value)
```
FeodorFitsner, line 2185: Shall we check that `value` is already in `self.__controls`?
...and I think deprecation notice should say "use `page.open(banner)` instead of adding to overlay?
The same applies to other popups.

sdk/python/packages/flet-core/src/flet_core/banner.py
```diff
@@ -16,31 +16,33 @@ class Banner(Control):
...
     import flet as ft
 
+
     def main(page):
+        page.horizontal_alignment = ft.CrossAxisAlignment.CENTER
+
         def close_banner(e):
-            page.banner.open = False
-            page.update()
+            page.close(banner)
```
FeodorFitsner, line 24: we can, actually, add `banner.close()` method, can we?

sdk/python/packages/flet-core/src/flet_core/page.py
ndonkoHenri: We could eventually... 
Any reason for that?

sdk/python/packages/flet-core/src/flet_core/banner.py
ndonkoHenri: If we are doing this, the next question will be, what about `banner.open()`, then `date_picker.pick_time()`... :)

FeodorFitsner (approved):

PR #3456, opened 2024-06-12
New property: `dropdown.Option.content`

FeodorFitsner:
Example:

```py
import flet as ft


def main(page: ft.Page):

    page.add(
        ft.Dropdown(
            width=200,
            options=[
                ft.dropdown.Option(
                    key="red",
                    content=ft.Row(
                        [
                            ft.Container(
                                width=24, height=24, bgcolor="red", border_radius=3
                            ),
                            ft.Text("Red"),
                        ]
                    ),
                ),
                ft.dropdown.Option("Green"),
                ft.dropdown.Option("Blue"),
            ],
        )
    )


ft.app(target=main)
```

Result:

image

image

## Summary by Sourcery

This pull request introduces a new `content` property for `dropdown.Option`, enabling the inclusion of custom widget content within dropdown options. The dropdown rendering logic has been updated to support this new feature, providing more flexibility in dropdown item presentation.

* **New Features**:
    - Introduced a new property `content` for `dropdown.Option` to allow custom widget content within dropdown options.
* **Enhancements**:
    - Updated the dropdown rendering logic to support custom content if the `content` property is provided, falling back to text otherwise.

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/dropdown.py
+++ b/sdk/python/packages/flet-core/src/flet_core/dropdown.py
@@ -1,5 +1,5 @@
 import time
-from typing import Any, Optional, Union, List
+from typing import Any, List, Optional, Union
 
 from flet_core.alignment import Alignment
 from flet_core.control import Control, OptionalNumber
@@ -23,6 +23,7 @@ class Option(Control):
         self,
         key: Optional[str] = None,
         text: Optional[str] = None,
+        content: Optional[Control] = None,
         alignment: Optional[Alignment] = None,
         text_style: Optional[TextStyle] = None,
         on_click=None,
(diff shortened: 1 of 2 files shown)


sdk/python/packages/flet-core/src/flet_core/dropdown.py
```diff
@@ -38,13 +39,21 @@ def __init__(
         assert key is not None or text is not None, "key or text must be specified"
```
ndonkoHenri, line 39: This assertion is needs to be updated.
We need to move it into `before_update()`, and check that either `key`, `text`, or `content` is provided.

Also, what's the difference between `text` and `key`? If there isn't, can we get rid of `key` in favour of `text`?

FeodorFitsner: Well, this assertion is kind of OK:
- If `content` is not provided then you have to provide either `key` or `text`.
- If `content` is provided you still need to provide `key` or/and `text` that will be used as a key.

Re: `key` and `text`:
- If `key` is not provided then `text` is used as a key.
- If `text` is not provided then `key` is used as a key.
- If both `key` and `text` are provided then `key` is used as a key and `text` as display value.

ndonkoHenri: Ah, alright makes sense.
But I think it needs to be moved to `before_update` so we can assert more often, instead of just once at instanciation.

ndonkoHenri (approved):

PR #3513, opened 2024-06-21
SearchView popup not to be opened on tap

InesaFitsner:
Currently, search View popup is opened on tap. Removed it so that user can call it from python instead.

## Summary by Sourcery

Removed the automatic opening of the search view on tap, allowing users to control this behavior from Python instead.


ndonkoHenri:
Don't you think it should be the default behavior?

FeodorFitsner:
I think it should be decided by developer on Python side.

ndonkoHenri:
> I think it should be decided by developer on Python side.

Ok, we could do it like that...

FeodorFitsner (approved):

Issue #3546, opened 2024-06-26
Web event type names differ from app event type names


brycepg:
### Duplicate Check

- [X] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

As a web app event names are minified or something:
$ python picharttest.py 
HOVER:  nified:a_W
HOVER:  nified:a_X
HOVER:  nified:a_T
HOVER:  nified:a_Y
HOVER:  nified:Jh
HOVER:  nified:a_X
HOVER:  nified:a_X

As a regular app event names are correct:
$ python picharttest.py 
HOVER:  PointerHoverEvent
HOVER:  PanDownEvent
HOVER:  TapDownEvent
HOVER:  TapUpEvent
HOVER:  PointerHoverEvent

### Code

```
import flet as ft

def main(page: ft.Page):
    normal_border = ft.BorderSide(0, ft.colors.with_opacity(0, ft.colors.WHITE))
    hovered_border = ft.BorderSide(6, ft.colors.WHITE)

    def on_chart_event(e: ft.PieChartEvent):
        for idx, section in enumerate(chart.sections):
            section.border_side = (
                hovered_border if idx == e.section_index else normal_border
            )
            if idx == e.section_index:
                print("HOVER: ", e.type)
        chart.update()

    chart = ft.PieChart(
        sections=[
            ft.PieChartSection(
                25,
                color=ft.colors.BLUE,
                radius=80,
                border_side=normal_border,
            ),
            ft.PieChartSection(
                25,
                color=ft.colors.YELLOW,
                radius=65,
                border_side=normal_border,
            ),
            ft.PieChartSection(
                25,
                color=ft.colors.PINK,
                radius=60,
                border_side=normal_border,
            ),
            ft.PieChartSection(
                25,
                color=ft.colors.GREEN,
                radius=70,
                border_side=normal_border,
            ),
        ],
        sections_space=1,
        center_space_radius=0,
        on_chart_event=on_chart_event,
        expand=True,
    )

    page.add(chart)

# ft.app(main)
ft.app(target=main, view=ft.AppView.WEB_BROWSER)
```

### To reproduce

1. Run code
2. Click on pi chart cell
3. Observe that the event type names have been minified? they don't match between web and regular app

### Expected behavior

Event type names should be preserved to preserve cross platform compatibility:

For example my code is looking for TapUpEvent which is renamed in the web app to `nified:a_Y`.

### Operating System

Linux

[...]

PR #3611, opened 2024-07-06
Fix `PieChartEvent.type` on web

ndonkoHenri:
Fixes https://github.com/flet-dev/flet/issues/3546

Tried another approach, but the issue is still present.

AI's opinion on the cause of the issue:
> When you run your Flutter application on the web, the Dart compiler minifies the code to reduce the size of the JavaScript output. This process often includes renaming classes and variables to shorter names (e.g., aj_) to optimize the code for performance and load times. As a result, PointerHoverEvent might be minified to something like aj_, making it harder to understand directly from the debug output.

What I found strange is the fact that, when testing on web using `flutter run -d chrome` it works (shows the right types).

## Summary by Sourcery

This pull request addresses the issue with `PieChartEvent.type` on web by mapping Flutter event types to readable event names. It also enhances the event data by adding local position information and introduces an enum for event types in the Python SDK.

- **Bug Fixes**:
    - Fixed the issue with `PieChartEvent.type` on web by mapping Flutter event types to readable event names.
- **Enhancements**:
    - Added `localPosition` to `PieChartEventData` to capture the local position of events.
    - Introduced `PieChartEventType` enum in Python to standardize event type values.

### DIFF
--- a/packages/flet/lib/src/controls/piechart.dart
+++ b/packages/flet/lib/src/controls/piechart.dart
@@ -15,15 +15,38 @@ import '../utils/borders.dart';
 import '../utils/text.dart';
 import 'create_control.dart';
 
+const eventMap = {
+  "FlPointerEnterEvent": "pointerEnter",
+  "FlPointerExitEvent": "pointerExit",
+  "FlPointerHoverEvent": "pointerHover",
+  "FlPanCancelEvent": "panCancel",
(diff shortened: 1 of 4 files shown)


FeodorFitsner:
Can you resolve conflict here please?

FeodorFitsner (approved):

PR #3659, opened 2024-07-17
improve type hint for OptionalEventCallable

zrr1999:
## Description

The current annotation is only correct for synchronous functions.  
I will replace `Callable[[ControlEvent], None]]` with `Callable[[ControlEvent], Union[None, Coroutine[None, None, None]]]` to support asynchronous functions.

## Test Code

```python
import flet as ft


async def demo(e):
    pass


ft.TextButton("Hello, World!", on_click=demo)
```

## Type of change

## Checklist:

- [x] I signed the CLA.
- [x] I have commented my code, particularly in hard-to-understand areas

- [x] New and existing tests pass locally with my changes
- [x] I have made corresponding changes to the [documentation](https://github.com/flet-dev/website) (if applicable)

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/types.py
+++ b/sdk/python/packages/flet-core/src/flet_core/types.py
@@ -1,5 +1,14 @@
 from enum import Enum, EnumMeta
-from typing import Any, Callable, Coroutine, Dict, Optional, Protocol, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Protocol,
+    Tuple,
+    TypeVar,
+    Union,
+)
 from warnings import warn
 
 from flet_core.animation import Animation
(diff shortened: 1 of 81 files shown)


ndonkoHenri:
Thanks for your contribution! You raised an important point.
I made a commit to further enhance the typing of event-handlers in all the files.

I also considered what the author of #3673 said concerning the return type: 
> flet probably does not care about the return value, and therefore should not place any restrictions on it.

sdk/python/packages/flet-core/src/flet_core/video.py
```diff
@@ -559,36 +560,36 @@ def on_enter_fullscreen(self):
         return self._get_event_handler("enter_fullscreen")
 
     @on_enter_fullscreen.setter
-    def on_enter_fullscreen(self, handler: OptionalEventCallable):
+    def on_enter_fullscreen(self, handler: DefaultOptionalEventCallable):
         self._add_event_handler("enter_fullscreen", handler)
```
FeodorFitsner, line 564: I think instead of `DefaultOptionalEventCallable` (which sounds kind of heavy) we could use just `EventCallable` everywhere. We don't have non-optional event handlers, right?

sdk/python/packages/flet-core/src/flet_core/types.py
```diff
@@ -393,7 +402,11 @@ class VisualDensity(Enum):
 
 
 # Events
-OptionalEventCallable = Optional[Callable[[ControlEvent], None]]
+ControlEventType = TypeVar("ControlEventType", bound=ControlEvent)
+EventCallable = Callable[[ControlEventType], Any]
+OptionalEventCallable = Optional[EventCallable]
+DefaultOptionalEventCallable = OptionalEventCallable[ControlEvent]
```
FeodorFitsner, line 408: How about:

```suggestion
OptionalControlEventCallable = OptionalEventCallable[ControlEvent]
```

FeodorFitsner (approved):

PR #3779, opened 2024-08-09
fix!: `InputFilter` clears `TextField` when an invalid character is entered 

### LINKED ISSUE #3769: Fix `InputFilter` clearing field when an invalid character is entered
### Discussed in https://github.com/flet-dev/flet/discussions/3764

<sup>Originally posted by **ClearSafety** August  5, 2024</sup>
The class ft.InputFilter makes the field content be erased if a not allowed charactered is typed by the user.
On the other hand, the class ft.NumbersOnlyInputFilter() does not do it, it simply does not allow the not allowed typed value been entered in the field.

Using ft.NumbersOnlyInputFilter(), if you type 'd', for instance, nothing happens.
Using ft.InputFilter(regex_string='^\d*(\.\d{0,2})?$'), if you type 'd', everything previously type is deleted.

I'd like to suggest that 'InputFilter' could have the same behaviour of NumbersOnlyInputFilter</div>

ndonkoHenri:
## Description
Fixes #3769

A custom `FilteringTextInputFormatter` has been created to prevent non-matching characters from causing the deletion of the whole field content.

## Breaking Change
Solving this issue resulted to a breaking change: most of the `regex_string`s must be updated to use anchors.

### Why?
This is caused by `RegExp.hasMatch()` allowing non-matching characters to be entered/visible (without replacement), as long as the field content contains a matching substring.

### Migration
Use start(`^`) and end(`$`) anchors in the pattern/regexString:
- `r"[a-zA-Z]"`   ->   `r"^[a-zA-Z]*$"`
- `r"[0-9]"`  ->  `r"^[0-9]*$"`

In the above cases, the `*` is important, else it will be impossible to delete the last character in the field.
 
#### AI Prompt
To ease migration, below is a simple prompt that can be useful:
```
Help me update the following regex pattern: ####### 
Ensure the entire string matches the pattern and allows for an empty string. 
```
## Test Code

```python
import flet as ft


def main(page: ft.Page):

    page.add(
        ft.TextField(
            label="Ex: 2244.89",
            keyboard_type=ft.KeyboardType.NUMBER,
            input_filter=ft.InputFilter(
                allow=True,
                regex_string=r"^\d*(\.\d{0,2})?$",
                replacement_string="",
            ),
        )
    )


ft.app(target=main)

```

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Summary by Sourcery

Fix issue where `InputFilter` clears `TextField` when an invalid character is entered by introducing `CustomFilteringTextInputFormatter`. This change requires updating most `regex_string`s to use start (`^`) and end (`$`) anchors to ensure proper matching.

Bug Fixes:
- Prevent `InputFilter` from clearing the entire `TextField` content when an invalid character is entered.

Enhancements:
- Introduce `CustomFilteringTextInputFormatter` to handle input filtering with more control over regex patterns and matching behavior.

### DIFF
--- a/sdk/python/packages/flet-core/src/flet_core/textfield.py
+++ b/sdk/python/packages/flet-core/src/flet_core/textfield.py
@@ -1,6 +1,5 @@
 import dataclasses
 import time
-from dataclasses import field
 from enum import Enum
 from typing import Any, Optional, Union, List
 
@@ -54,18 +53,22 @@ class TextCapitalization(Enum):
 @dataclasses.dataclass
 class InputFilter:
     regex_string: str
-    allow: bool = field(default=True)
-    replacement_string: str = field(default="")
+    allow: bool = True
+    replacement_string: str = ""
+    multiline: bool = False
+    case_sensitive: bool = True
+    unicode: bool = False
+    dot_all: bool = False
 
 
 class NumbersOnlyInputFilter(InputFilter):
     def __init__(self):
-        super().__init__(regex_string=r"[0-9]")
+        super().__init__(regex_string=r"^[0-9]*$", allow=True, replacement_string="")
 
 
 class TextOnlyInputFilter(InputFilter):
     def __init__(self):
-        super().__init__(regex_string=r"[a-zA-Z]")
+        super().__init__(regex_string=r"^[a-zA-Z]*$", allow=True, replacement_string="")
 
 
 class TextField(FormFieldControl, AdaptiveControl):
@@ -130,7 +133,7 @@ class TextField(FormFieldControl, AdaptiveControl):
         on_focus: OptionalEventCallable = None,
         on_blur: OptionalEventCallable = None,
         #
-        # FormField specific
+        # FormField
         #
         text_size: OptionalNumber = None,
         text_style: Optional[TextStyle] = None,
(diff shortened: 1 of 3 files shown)


sdk/python/packages/flet-core/src/flet_core/textfield.py
```diff
@@ -54,18 +53,22 @@ class TextCapitalization(Enum):
...
 class InputFilter:
     regex_string: str
-    allow: bool = field(default=True)
-    replacement_string: str = field(default="")
+    allow: bool = True
+    replacement_string: str = ""
+    multiline: bool = False
+    case_sensitive: bool = True
+    unicode: bool = False
+    dot_all: bool = False
```
FeodorFitsner, line 61: wouldn't be more "handy" to have `unicode = True` and `case_sensitive = False` by default?

ndonkoHenri: Those are actually the default flutter values of `RegExp`.
What's your idea?

FeodorFitsner: Apparently, unicode mode is not what I thought: https://api.flutter.dev/flutter/dart-core/RegExp/isUnicode.html
OK, let's keep defaults as they are in RegExp.

FeodorFitsner (approved):

PR #4180, opened 2024-10-17
Migrate `colors` and `icons` variables to Enums

### LINKED ISSUE #4169: Migrate `colors` and `icons` to Enums
### Discussed in https://github.com/flet-dev/flet/discussions/4160

<sup>Originally posted by **Kenshin9977** October 14, 2024</sup>
While working on a project using flet I saw that any import made from flet took around 5 seconds. While investigating this I found out that 3 of those 5 seconds were used on `import icons` in flet_core/icons.py.
The file defines ~8000 global variables in this fashion:

```
TEN_K = "ten_k"
TEN_K_SHARP = "ten_k_sharp"
TEN_K_ROUNDED = "ten_k_rounded"
TEN_K_OUTLINED = "ten_k_outlined"
TEN_MP = "ten_mp"
...
ZOOM_OUT_MAP = "zoom_out_map"
ZOOM_OUT_MAP_SHARP = "zoom_out_map_sharp"
ZOOM_OUT_MAP_ROUNDED = "zoom_out_map_rounded"
ZOOM_OUT_MAP_OUTLINED = "zoom_out_map_outlined"

icons_list = [v for k, v in vars().items() if k.isupper()]
```

I tested removing all of those lines and it effectively more than halved the time it takes to import flet and consequently my app's launch time. Of course, removing all these values isn't an option but using an enum, a dict, a seperate JSON or even put those in several optional packages are ideas to radically improve flet's import time.</div>

ndonkoHenri:
## Description

Fixes #4169

## Test Code
```python
import flet as ft

print(ft.colors.BLACK)
print(ft.colors.RED.with_opacity(0.5))
print(ft.colors.random())

print(ft.icons.WB_SUNNY)
print(ft.icons.random())


# deprecated - use random() instead
print(ft.icons.random_icon())
print(ft.colors.random_color())


# broken
print(ft.colors.with_opacity(0.5, "blue"))
``` 

## Note
- I named the enums colors and icons (lowercased) mainly for backward compatibility, however the python convention is to have class names in CapWords. We could actually have both then deprecate the lowercase.
-  Once we conclude with the right way to do this, I will proceed with updating the cupertino icons/colors then the controls with color-props.


FeodorFitsner:
I'm not sure about `MaterialIcons` and `MaterialColors`. Material is the default theme in Flet. I like just `Colors` and `Icons` more.

FeodorFitsner:
...and cupertino is rarely used, so let's leave `Colors` and `Icons`.

FeodorFitsner (approved):

PR #4374, opened 2024-11-14
feat: magic methods in Controls (ex: `__contains__`)

### LINKED ISSUE #4217: Implement useful magic methods in Controls (ex: `__contains__`)
### Duplicate Check

I would like that in such elements as row, columns and so on it would be possible:
1. add an element as in `page.add()
2. use `__contains__ `method
```python
column = ft.Column()

column.add(
    text
)

print(text in column)
```

### Suggest a solution

```python
# add element in column
def add(self, control):
    self.controls.append(control)
    self.update()
```

### Screenshots

_No response_

### Additional details

This is a good way to bring the library to the same syntax, instead of using `page.add()` or `column.controls.append()`.

ndonkoHenri:
Fixes #4217

## Summary by Sourcery

New Features:
- Introduce `__contains__` magic method to various control classes, allowing for more intuitive membership testing.

### DIFF
--- a/sdk/python/packages/flet/src/flet/core/column.py
+++ b/sdk/python/packages/flet/src/flet/core/column.py
@@ -159,6 +159,9 @@ class Column(ConstrainedControl, ScrollableControl, AdaptiveControl):
     def _get_children(self):
         return self.__controls
 
+    def __contains__(self, item):
+        return item in self.__controls
+
     # Public methods
     def clean(self):
         super().clean()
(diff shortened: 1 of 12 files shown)


ndonkoHenri:
Let me know if I missed a control.
Also, should we actually check in all the props of type `Control` or just the `controls` is sufficient ?

FeodorFitsner (approved):

PR #4389, opened 2024-11-17
chore: reorganize `build` cli

ndonkoHenri:
## Summary by Sourcery

Enhancements:
- Refactor the build command in the Flet CLI to improve code organization and readability by splitting functionalities into smaller methods.


sdk/python/packages/flet-cli/src/flet_cli/commands/build.py
```diff
@@ -49,28 +47,6 @@ class Command(BaseCommand):
...
-        self.python_module_name = None
-        self.get_pyproject = None
-        self.python_app_path = None
-        self.no_rich_output = None
-        self.emojis = {}
-        self.dart_exe = None
-        self.verbose = False
-        self.build_dir = None
-        self.flutter_dir: Optional[Path] = None
-        self.flutter_exe = None
```
ndonkoHenri: why remove this?

FeodorFitsner: It makes type checker really "red":

image

ndonkoHenri: The red lines you are showing seem to come from using `self.options.X` instead? 

Without that block of initializations i get: 
image

SO discussion: https://stackoverflow.com/a/19292653 (common python practice)

FeodorFitsner: Looks like it's coming from `self.options`. What would be the solution?

image

ndonkoHenri: Perhaps use `cast` as you did in the others? 
Or perhaps just mute the warning?

PR #5265, opened 2025-05-01
v1: patches + improvements

ndonkoHenri:
## Summary by Sourcery

Refactor and improve storage paths, responsive design, and control handling across Dart and Python implementations of the Flet framework

New Features:
- Added ResponsiveRowBreakpoint enum in Python
- Introduced platform-specific method checks for storage paths

Enhancements:
- Redesigned StoragePaths service to use invoke method pattern
- Improved responsive row breakpoint handling
- Simplified control and shader mask implementations
- Added platform-specific method checks in Python implementation

Chores:
- Cleaned up unused imports
- Simplified error handling and control logic

### DIFF
--- a/packages/flet/lib/src/controls/cupertino_timer_picker.dart
+++ b/packages/flet/lib/src/controls/cupertino_timer_picker.dart
@@ -26,15 +26,19 @@ class _CupertinoTimerPickerControlState
     Widget picker = CupertinoTimerPicker(
       mode: widget.control
           .getCupertinoTimerPickerMode("mode", CupertinoTimerPickerMode.hms)!,
-      initialTimerDuration: widget.control.getDuration("value", Duration.zero)!,
+      initialTimerDuration: widget.control
+          .getDuration("value", Duration.zero, DurationUnit.seconds)!,
       minuteInterval: widget.control.getInt("minute_interval", 1)!,
       secondInterval: widget.control.getInt("second_interval", 1)!,
       itemExtent: widget.control.getDouble("item_extent", 32.0)!,
       alignment: widget.control.getAlignment("alignment", Alignment.center)!,
       backgroundColor: widget.control.getColor("bgcolor", context),
       onTimerDurationChanged: (duration) {
-        widget.control.updateProperties({"value": duration});
-        widget.control.triggerEvent("change", duration);
+        final valueIsInt = widget.control.get("value") is int;
+        // preserve the value's type
+        final d = valueIsInt ? duration.inSeconds : duration;
+        widget.control.updateProperties({"value": d});
+        widget.control.triggerEvent("change", d);
       },
     );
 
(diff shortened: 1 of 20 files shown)


packages/flet/lib/src/services/url_launcher.dart
```diff
@@ -26,9 +27,9 @@ class UrlLauncherService extends FletService {
       case "launch_url":
         return openWebBrowser(args["url"]!,
             webWindowName: args["web_window_name"],
-            webPopupWindow: args["web_popup_window"],
-            windowWidth: args["window_width"],
-            windowHeight: args["window_height"]);
+            webPopupWindow: parseBool(args["web_popup_window"]),
```
FeodorFitsner, line 30: I don't think we need `parse` here as args are incoming with their right type.

ndonkoHenri: You know python doesn't check types. So if a user sends a float instead of an int it will fail. 
Parsing to avoid unnecessary possible errors - doesn't cost us anything.

sdk/python/packages/flet/src/flet/controls/core/shader_mask.py
```diff
@@ -49,6 +47,6 @@ def main(page: ft.Page):
     """
 
     shader: Gradient
-    content: Optional[Control] = None
-    blend_mode: Optional[BlendMode] = None
+    content: OptionalControl = None
```
FeodorFitsner, line 50: I'm still not completely sold we should introduce a lot of "specific" types, like `OptionalControl` where you have to go there and see what that type really means. Moreover, starting from Python 3.11 it has nicer syntax: `Control | None`, so eventually we won't need that at all.

packages/flet/lib/src/services/url_launcher.dart
ndonkoHenri: When we integrate pydantic or anything similar, we could then remove these, as we would be 100% sure incoming type will be correct.

packages/flet/lib/src/controls/cupertino_timer_picker.dart
```diff
@@ -26,15 +26,19 @@ class _CupertinoTimerPickerControlState
...
       secondInterval: widget.control.getInt("second_interval", 1)!,
       itemExtent: widget.control.getDouble("item_extent", 32.0)!,
       alignment: widget.control.getAlignment("alignment", Alignment.center)!,
       backgroundColor: widget.control.getColor("bgcolor", context),
       onTimerDurationChanged: (duration) {
-        widget.control.updateProperties({"value": duration});
-        widget.control.triggerEvent("change", duration);
+        final valueIsInt = widget.control.get("value") is int;
+        // preserve the value's type
+        final d = valueIsInt ? duration.inSeconds : duration;
```
FeodorFitsner, line 39:
```suggestion
        final d = widget.control.get("value") is int ? duration.inSeconds : duration;
```

FeodorFitsner (approved):

PR #5342, opened 2025-05-29
v1: Declarative/Reactive approach in Flet

FeodorFitsner:
This PR introduces a **declarative/reactive approach** to building UIs in Flet. Developers can now mix **imperative** and **declarative** patterns in the same app, enabling more flexible and functional UI code.

## Declarative "Counter" example

```py
from dataclasses import dataclass

import flet as ft


@dataclass
class AppState:
    count: int

    def increment(self):
        self.count += 1


def main(page: ft.Page):
    state = AppState(count=0)

    page.floating_action_button = ft.FloatingActionButton(
        icon=ft.Icons.ADD, on_click=state.increment
    )
    page.add(
        ft.ControlBuilder(
            state,
            lambda state: ft.SafeArea(
                ft.Center(ft.Text(value=f"{state.count}", size=50)),
                expand=True,
            ),
            expand=True,
        )
    )


ft.run(main)
```

## Declarative Form example

```py
from dataclasses import dataclass
from typing import cast

import flet as ft

@dataclass
class Form:
    first_name: str = ""
    last_name: str = ""

    def set_first_name(self, value):
        self.first_name = value

    def set_last_name(self, value):
        self.last_name = value

    async def submit(self, e: ft.ControlEvent):
        e.page.show_dialog(
            ft.AlertDialog(
                title="Hello",
                content=ft.Text(f"{self.first_name} {self.last_name}!"),
            )
        )

    async def reset(self):
        self.first_name = ""
        self.last_name = ""


def main(page: ft.Page):
    form = Form()

    page.add(
        ft.ControlBuilder(
            form,
            lambda state: ft.Column(
                cast(
                    list[ft.Control],
                    [
                        ft.TextField(
                            label="First name",
                            value=form.first_name,
                            on_change=lambda e: form.set_first_name(e.control.value),
                        ),
                        ft.TextField(
                            label="Last name",
                            value=form.last_name,
                            on_change=lambda e: form.set_last_name(e.control.value),
                        ),
                        ft.Row(
                            [
                                ft.FilledButton("Submit", on_click=form.submit),
                                ft.FilledTonalButton("Reset", on_click=form.reset),
                            ]
                        ),
                    ],
                )
            ),
        )
    )

ft.run(main)
```

### Frozen controls

Frozen controls are **immutable** — they must be re-created via constructors. Their properties cannot be updated after creation. Flet uses **deep structural comparison** to detect changes in frozen controls and determine whether they need to be re-rendered.

### `@data_view` decorator

Introduces a "UI as a function of state" paradigm.

A function decorated with `@data_view` accepts one or more inputs (state/data) and returns a Control or a list of Controls. The returned controls are frozen and eligible for diffing and caching.

Example:

```py
@dataclass
class User:
    id: int
    name: str
    age: int
    verified: bool


users = [
    User(1, "John Smith", 20, True),
    User(2, "Alice Wong", 32, True),
    User(3, "Bob Bar", 40, False),
]

@ft.data_view
def user_details(user: User):
    return ft.Card(
        ft.Column(
            [
                ft.Text(f"Name: {user.name}"),
                ft.Text(f"Age: {user.age}"),
                ft.Checkbox(label="Verified", value=user.verified),
            ]
        ),
        list_key=user.id,
    )

@ft.data_view
def users_list(users):
    return ft.Column([user_details(user) for user in users])
```

When the underlying `users` list changes, the developer is responsible for **re-invoking** the `users_list()` function to generate a new control tree:

```py
page.add(users_list(users))

# Later...

users.append(User(4, name="Someone Else", age=99, verified=False))
page.controls[0] = users_list(users)

del users[1]
page.controls[0] = users_list(users)
```

### `key` property

The `key` Control property helps Flet **identify and match items across control lists** when comparing old and new frozen control trees. If two controls in the same list share the same `key`, Flet performs a deep comparison to detect changes. This improves performance and preserves state during updates.

If `key` is omitted or `None`, comparison is done by position and structure.

The `key` can be an instance of `ft.ValueKey`, `ft.ScrollKey`, or a literal value such as a string, number, or boolean.

On the Dart side, a `ft.ValueKey` or literal is assigned to the widget’s `key` property using a `ValueKey`. A `ft.ScrollKey`, on the other hand, is assigned using a `GlobalKey`.

### `ControlBuilder` control

`ControlBuilder` is a lightweight reactive control that rebuilds its content every time the UI is updated.

Constructor:

```py
ControlBuilder(state, builder, ...)
```

* `state`: any data object.
* `builder`: a function that takes `state` and returns a `Control`.

See Counter example above.

### Event handlers without `e`

Simple event handlers can now omit `e` parameter, for example both of these work:

```py
button_1.on_click = lambda: print("Clicked!")
button_2.on_click = lambda e: print("Clicked!", e)
```

or

```py
def increment():
   print("Increment clicked")

inc_btn.on_click = increment
```

### `Control.before_event(e)` hook

The method is called before calling any event handler.

It receives an instance of `ControlEvent` as parameter and should return either `True` or `False`. Returning `False` cancels event handler. Example implementation in `Page` class:

```py
    def before_event(self, e: ControlEvent):
        if isinstance(e, RouteChangeEvent):
            if self.route == e.route:
                return False
            self.query()
        return super().before_event(e)
``` 

### `ft.context.page` works in all event handlers

It's now possible to get a reference to a current `Page` instance in any part of Flet program.

### DIFF
--- a/sdk/python/packages/flet/src/flet/controls/base_control.py
+++ b/sdk/python/packages/flet/src/flet/controls/base_control.py
@@ -190,7 +190,8 @@ class BaseControl:
 
     # public methods
     def update(self) -> None:
-        # TODO - frozen control cannot be updated
+        if hasattr(self, "_frozen"):
+            raise Exception("Frozen control cannot be updated.")
         assert self.page, (
             f"{self.__class__.__qualname__} Control must be added to the page first"
         )
--- a/sdk/python/packages/flet/src/flet/controls/control.py
+++ b/sdk/python/packages/flet/src/flet/controls/control.py
@@ -6,10 +6,7 @@ from flet.controls.material.badge import BadgeValue
 from flet.controls.material.tooltip import TooltipValue
 from flet.controls.types import Number, ResponsiveNumber
 
-__all__ = [
-    "Control",
-    "OptionalControl"
-]
+__all__ = ["Control", "OptionalControl"]
 
 
 @dataclass(kw_only=True)
@@ -17,7 +14,7 @@ class Control(BaseControl):
     """
     TBD
     """
-    
+
     expand: Optional[Union[bool, int]] = None
     """
     When a child Control is placed into a [`Column`](https://flet.dev/docs/controls/column) 
@@ -180,7 +177,7 @@ class Control(BaseControl):
 
     | Breakpoint | Dimension |
     |---|---|
-    | xs | \<576px |
+    | xs | <576px |
     | sm | ≥576px |
     | md | ≥768px |
     | lg | ≥992px |
@@ -189,7 +186,7 @@ class Control(BaseControl):
 
     If `col` property is not specified, it spans the maximum number of columns (12).
     """
-    
+
     opacity: Number = 1.0
     """
     Defines the transparency of the control.
@@ -217,7 +214,7 @@ class Control(BaseControl):
     all its children if any) from rendering on a page canvas. Hidden controls cannot be 
     focused or selected with a keyboard or mouse and they do not emit any events.
     """
-    
+
     disabled: bool = False
     """
     Every control has `disabled` property which is `False` by default - control and all 
@@ -247,12 +244,12 @@ class Control(BaseControl):
 
     def before_update(self):
         super().before_update()
-        assert (
-            0.0 <= self.opacity <= 1.0
-        ), "opacity must be between 0.0 and 1.0 inclusive"
-        assert self.expand is None or isinstance(
-            self.expand, (bool, int)
-        ), "expand must be of bool or int type"
+        assert 0.0 <= self.opacity <= 1.0, (
+            "opacity must be between 0.0 and 1.0 inclusive"
+        )
+        assert self.expand is None or isinstance(self.expand, (bool, int)), (
+            "expand must be of bool or int type"
+        )
 
     def clean(self) -> None:
         raise Exception("Deprecated!")
--- a/sdk/python/packages/flet/src/flet/messaging/session.py
+++ b/sdk/python/packages/flet/src/flet/messaging/session.py
@@ -231,7 +231,7 @@ class Session:
 
     async def auto_update(self, control: BaseControl):
         while control:
-            if control.is_isolated():
+            if control.is_isolated() and not hasattr(control, "_frozen"):
                 control.update()
                 break
             control = control.parent


sdk/python/packages/flet/src/flet/controls/types.py
```diff
@@ -358,7 +358,12 @@ class LocaleConfiguration:
 # Events
 EventType = TypeVar("EventType", bound=ControlEvent)
 OptionalEventCallable = Optional[Callable[[EventType], Any]]
-OptionalControlEventCallable = Optional[Callable[[ControlEvent], Any]]
+OptionalControlEventCallable = Optional[
+    Union[
+        Callable[[], Any],
+        Callable[[ControlEvent], Any],
+    ]
+]
```
ndonkoHenri, line 366: I love the idea. Any reason for not having thesame in `OptionalEventCallable`?

sdk/python/packages/flet/src/flet/controls/core/center.py
```diff
@@ -0,0 +1,19 @@
...
+class Center(Control):
+    """
+    Centers content.
+
+    -----
+
+    Online docs: https://flet.dev/docs/controls/center
+    """
+
+    content: Optional[Control] = None
```
ndonkoHenri, line 19: We need to discuss on how such "utility" controls should be implemented. 

Till date, we have been having them implicitly, for example: `Container` has `padding`, `margin` props; and with this PR you are willing to introduce the explicit approach (similar to Flutter): everything is a widget/control. 

I however see disadvantages in both styles: 
- implicit: Flutter has alot of utility widgets. Having all of them in each Controls will definitely not be a good idea.
- explicit: kind of breaks the "Flet style" we had till now. What will happen to guys like border, alignment etc? Are we going to convert them to Controls too? (i guess, no 🥹)

I actually spent some time some months ago thinking of a nice approach concerning this issue, and came up with a [`UtilityControl`](https://github.com/flet-dev/flet/commit/14ea6f5a7cca386990d3ce36d4fb5ee434e363d1), which should have all this utility props (padding, margin, border-radius, etc) - looking at the code you will get the idea. 

Let me have your thoughts on all this.

sdk/python/packages/flet/src/flet/controls/types.py
FeodorFitsner: Sure, added.

sdk/python/packages/flet/src/flet/controls/base_control.py
```diff
@@ -95,9 +95,10 @@ def new_post_init(self: T, *args):
 
 @dataclass(kw_only=True)
 class BaseControl:
-    _i: int = field(init=False)
+    _i: int = field(init=False, compare=False)
     _c: str = field(init=False)
     data: Any = skip_field()
```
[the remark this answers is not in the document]
FeodorFitsner: So, I think we should not use `compare` at all and rely only on metadata.

We need two "behaviors" for our fields:
- a flag that tells whether field value should be sent to Dart, i.e. included into serialized control data. This is `skip` right now.
- a flag that tells whether field must be excluded from comparison of two "frozen" controls. It's currently only `Control._i` that must be excluded from comparison, but sent to Dart.

Fixed here: https://github.com/flet-dev/flet/commit/7d49afa5eaa39bda3e8942181cb3d089c388463f

PR #5402, opened 2025-06-25
v1: replace `:::note` by `Note:` for mkdocs

ndonkoHenri:
With `:::note`, mkdocs thinks it is supposed to import a package/class, thereby failing to build+deploy.
Needed for flet-charts/MatplotlibChart, which inherits from `Container`.

## Summary by Sourcery

Standardize internal type representation and update docstrings

Enhancements:
- Replace public `type` fields with private `_type` in dataclass models for shapes, gradients, painting, borders, and controls
- Update parsing functions to use `_type` and introduce default fallback for gradient parsing
- Refactor assertion formatting in CupertinoDatePicker for consistent style

Documentation:
- Convert mkdocs `:::note` and `:::info` admonitions to inline `Note:` and `Info:` blocks in docstrings

### DIFF
--- a/sdk/python/packages/flet/src/flet/controls/cupertino/cupertino_date_picker.py
+++ b/sdk/python/packages/flet/src/flet/controls/cupertino/cupertino_date_picker.py
@@ -143,13 +143,12 @@ class CupertinoDatePicker(ConstrainedControl):
   
     Value is of type [`CupertinoDatePickerDateOrder`](https://flet.dev/docs/reference/types/cupertinodatepickerdateorder).
   
-    :::note
-    The final order in which the columns are displayed is also influenced by 
-    the `date_picker_mode`. For example, when using 
-    `date_picker_mode=CupertinoDatePickerMode.MONTH_YEAR`, 
-    both `CupertinoDatePickerDateOrder.DAY_MONTH_YEAR` and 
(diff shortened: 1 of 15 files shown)


sdk/python/packages/flet/src/flet/controls/cupertino/cupertino_date_picker.py
```diff
@@ -143,13 +143,12 @@ class CupertinoDatePicker(ConstrainedControl):
...
-    both `CupertinoDatePickerDateOrder.DAY_MONTH_YEAR` and 
-    `CupertinoDatePickerDateOrder.MONTH_DAY_YEAR` will result in the month|year order.
-    :::
+    Note:
+        The final order in which the columns are displayed is also influenced by 
+        the `date_picker_mode`. For example, when using 
+        `date_picker_mode=CupertinoDatePickerMode.MONTH_YEAR`, 
+        both `CupertinoDatePickerDateOrder.DAY_MONTH_YEAR` and 
+        `CupertinoDatePickerDateOrder.MONTH_DAY_YEAR` will result in the month|year order.
     """
```
FeodorFitsner, line 152: We are going to use Docusaurus for built-in docs and while this works for mkdocs (which we are not yet decided to to use for built-in docs for now) it will break a note in Docusaurus.

FeodorFitsner (approved):

PR #5466, opened 2025-07-21
v1: Documentation with mkdocs

ndonkoHenri:


sdk/python/packages/flet/src/flet/controls/buttons.py
```diff
@@ -47,72 +63,132 @@ class OutlinedBorder:
...
+        Returns a copy of this object with the specified properties overridden.
+        """
+        return ContinuousRectangleBorder(
+            side=side if side is not None else self.side,
+            radius=radius if radius is not None else self.radius,
+        )
+
+
+CircleBorder()
+
```
FeodorFitsner: typo?

packages/flet/lib/src/controls/chip.dart
```diff
@@ -87,15 +89,16 @@ class _ChipControlState extends State<ChipControl> {
...
+          widget.control.getTextStyle("label_text_style", Theme.of(context)),
       selectedColor: widget.control.getColor("selected_color", context),
       selectedShadowColor:
           widget.control.getColor("selected_shadow_color", context),
       shadowColor: widget.control.getColor("shadow_color", context),
       shape: widget.control.getShape("shape", Theme.of(context)),
       color: widget.control.getWidgetStateColor("color", Theme.of(context)),
       surfaceTintColor: widget.control.getColor("surface_tint_color", context),
-      pressElevation: widget.control.getDouble("click_elevation"),
+      pressElevation: widget.control.getDouble("elevation_on_click"),
```
FeodorFitsner, line 101: Let's use the same property name as in Flutter?

FeodorFitsner:
Some design notes:

"attr" labels in front of properties look a bit "aggressive":

image

I would put them behind property name and maybe renamed to `property` in more light color, like Pydantic did:

image

so, when I glance at the page I first see property name and not "attr".

The same for methods.

FeodorFitsner:
Canvas examples are still at the bottom.

FeodorFitsner:
I'm wondering is it possible to take class description and output it at the top of the page?
Examples right after page title, without any into, look a bit out of place.

image

This is how it looks now:

image

packages/flet/lib/src/controls/chip.dart
ndonkoHenri: not really. 
onPress -> on_click

ndonkoHenri:
> I'm wondering is it possible to take class description and output it at the top of the page?

Tried `::: flet.Switch.__doc__` but it made mkdocs break. I asked: https://github.com/mkdocstrings/python/discussions/302

PR #5545, opened 2025-08-13
change base64.py example to src_base64.py to avoid shadowing built in…

OwenMcDonnell:
Change base64.py example to `src_base64.py` to avoid shadowing built in base64 module

## Description

When trying to run existing image control examples in folder with one example named `base64.py` the import of python's base64 module is shadowed


FeodorFitsner:
How do you import it?

OwenMcDonnell:
You mean the base64 module? It's part of Python library. So when you try to run the example from a directory with a file named base64.py, that file gets loaded instead of the python library one.

FeodorFitsner:
Ah, OK.

PR #5558, opened 2025-08-16
v1(ci): use uv inline-scripting, prevent concurrent runs in `macos-integration-tests` GHA

ndonkoHenri:
## Summary by Sourcery

Switch CI pipelines to use uv's inline scripting and add concurrency safeguards to the macOS integration workflow

Enhancements:
- Switch CI uv run/build commands to inline scripting by removing explicit --with plugin flags in AppVeyor scripts
- Configure GitHub Actions macOS integration tests to only trigger on relevant path changes and cancel in-progress runs for the same branch
- Simplify Python-based CI helper scripts by removing redundant file mode arguments when opening files


.github/workflows/macos-integration-tests.yml
```diff
@@ -2,7 +2,17 @@ name: macOS Integration Tests
 
 on:
   push:
+    paths:
+      - 'sdk/python/packages/flet/src/**'
+      - 'sdk/python/packages/flet/integration_tests/**'
   pull_request:
```
FeodorFitsner, line 10: Khm, what if Dart part is updated?

.github/workflows/macos-integration-tests.yml
```diff
@@ -2,7 +2,17 @@ name: macOS Integration Tests
...
+    paths:
+      - 'sdk/python/packages/flet/src/**'
+      - 'sdk/python/packages/flet/integration_tests/**'
+
+concurrency:
+  group: macos-integration-tests-${{ github.ref }}
```
FeodorFitsner: What does `github.ref` mean here?

ndonkoHenri: > `${{ github.ref }}` is a GitHub Actions context variable that contains the full ref (reference) of the branch or tag that triggered the workflow. For example, for a branch it could be `refs/heads/v1-improve-gha`, and for a tag it could be `refs/tags/v1.0.0`. This is used to group workflow runs by branch or tag for concurrency control.

FeodorFitsner: OK, so it's per branch. Understood.

ndonkoHenri: just from modifying the group name again to prevent duplicate runs when pushing a commit to a branch which has an opened PR.
image

Now in such cases, only the PR run will take the lead, and the branch run will be cancelled.

Issue #5568, opened 2025-08-18
v1: Tooltip bgcolor has no effect


oussama-gourari:
### Duplicate Check

- [x] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

Specifying a `bgcolor` for a `ft.Tooltip` seems to have no effect.

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft


def main(page: ft.Page):
    page.add(
        ft.Text(
            value="Hover to see the tooltip",
            tooltip=ft.Tooltip(
                message="This is a tooltip with green bgcolor",
                padding=20,
                text_style=ft.TextStyle(size=20, color=ft.Colors.WHITE),
                bgcolor=ft.Colors.GREEN,
            ),
        ),
    )


ft.run(main)

```

</details>

### To reproduce

Run the sample code.

### Screenshots / Videos

<details open>
<summary>Captures</summary>

<img width="1366" height="730" alt="Image" src="https://github.com/user-attachments/assets/f4ff6761-452a-4bb8-9612-4a361884f561" />

</details>

### Operating System

Linux

### Operating system details

Manjaro Linux x86_64

### Flet version

0.70.0.dev5463

### Regression

I'm not sure / I don't know

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

7576457:
@FeodorFitsner 
@ndonkoHenri

Can I take this on?

ndonkoHenri:
@Klu1d sure! 👍🏾

PR #5575, opened 2025-08-19
fix: apply default Tooltip decoration values

7576457:
## Description

This PR addresses this issue -> #5568 

The original problem was that when creating a `BoxDecoration` for a `Tooltip`, default values were being set, but they were not saved anywhere. 
image

After I started saving it in a separate variable and passing it to the `decoration` attribute, I was able to get the colors specified in `tooltip.decoration`, but I still couldn’t get the `tooltip.bgcolor`.

The solution was to create a default value, like `defaultBoxDecoration`, and provide it to `parseBoxDecoration`.
After that, I noticed that the key being parsed from Flet is named `color`, even though in the Flet widget, the background color property is called `bgcolor`.

## Test Code

```python
import flet as ft


def main(page: ft.Page):
    page.add(
        ft.Text(
            value="Hover to see the tooltip",
            tooltip=ft.Tooltip(
                message="This is a tooltip with green bgcolor",
                padding=20,
                text_style=ft.TextStyle(size=20, color=ft.Colors.WHITE),
                bgcolor=ft.Colors.GREEN,
            ),
        ),
    )


ft.run(main)

```

## Type of change

## Checklist

- [x] I signed the CLA.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing tests pass locally with my changes
- [ ] I have made corresponding changes to the [documentation](https://github.com/flet-dev/website) (if applicable)

## Screenshots
изображение

## Summary by Sourcery

Ensure Tooltip decorations use proper default values by introducing and passing a default BoxDecoration and updating the parser to honor fallback properties.

Bug Fixes:
- Restore default border radius and background color for Tooltip decorations by creating and applying a default BoxDecoration.
- Fix parsing of background color using the "bgcolor" key instead of incorrectly using "color".

Enhancements:
- Extend parseBoxDecoration to accept a default BoxDecoration and use its borderRadius and color when parsing values.

### DIFF
--- a/packages/flet/lib/src/utils/box.dart
+++ b/packages/flet/lib/src/utils/box.dart
@@ -63,7 +63,7 @@ BoxDecoration? parseBoxDecoration(dynamic value, BuildContext context,
 
   var shape = parseBoxShape(value["shape"], BoxShape.rectangle)!;
   var borderRadius = parseBorderRadius(value["border_radius"]);
-  var color = parseColor(value["color"], theme);
+  var color = parseColor(value["bgcolor"], theme);
   var gradient = parseGradient(value["gradient"], theme);
   var blendMode = parseBlendMode(value["blend_mode"]);
 
--- a/packages/flet/lib/src/utils/tooltip.dart
+++ b/packages/flet/lib/src/utils/tooltip.dart
@@ -34,14 +34,25 @@ Tooltip? parseTooltip(dynamic value, BuildContext context, Widget widget) {
 
   /// The tooltip shape defaults to a rounded rectangle with a border radius of
   /// 4.0. Tooltips will also default to an opacity of 90%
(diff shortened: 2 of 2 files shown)


7576457:
> Let's not modify the `utils/box.dart` for now:

Your rough code works fine as a solution.
However, changes are still needed. In the file `utils/box.dart`, line 66:

```dart
var color = parseColor(value["color"], theme);
```

tries to get a color from the `value["decoration"]` dictionary. But there is no `color` key there — only `bgcolor`, since in Flet the `ft.BoxDecoration` attribute only has `bgcolor`.

Changing `color` to `bgcolor` is the correct and consistent solution. Without this change, it’s not possible to modify the color using `ft.BoxDecoration`, because it affects not only the `decoration` in `Tooltip`, but also `foreground_decoration` in `Container`.

PR #5669, opened 2025-09-30
ci: Include source-distribution when publishing Flet packages

ndonkoHenri:


FeodorFitsner:
Every platform-specific artifact here https://github.com/flet-dev/flet/actions/runs/18144469499 contains a source package for `flet_desktop` with platform-specific zip inside:

image


Source package *should not contain* any binaries in it, be small (only sources) and built only once perhaps in [Build flet and flet-cli Python packages](https://github.com/flet-dev/flet/actions/runs/18144469499/job/51643094275#logs) job.

ndonkoHenri:
> Source package should not contain any binaries in it, be small (only sources) and built only once

Solved: https://github.com/flet-dev/flet/actions/runs/18176278704/job/51744130968?pr=5669#step:4:24

FeodorFitsner:
> > Source package should not contain any binaries in it, be small (only sources) and built only once
> 
> Solved: https://github.com/flet-dev/flet/actions/runs/18176278704/job/51744130968?pr=5669#step:4:24

`flet_desktop-{ver}.tar.gz` is fine.

We need `flet_desktop_light-{ver}.tar.gz` too, right?

And `flet_web-{ver}.tar.gz` should be small as well.

PR #5690, opened 2025-10-08
v1: Migrate Flet extensions to a main repo

FeodorFitsner:


7576457:
Hold on, if I remember right, the plan was to take them out.
Did you decide to bring the packages back in?

FeodorFitsner:
> Hold on, if I remember right, the plan was to take them out. Did you decide to bring the packages back in?

That's correct, that was a plan. We decided to bring our own "built-in" extensions back to a main repo, for now - as the core is still unstable - to reduce the amount of boilerplate and simplify maintenance. Truly external extensions - 3rd-party or those not pre-compiled into a platform client inside the wheel - will be in separate repositories.

PR #5742, opened 2025-10-23
Handle non-event `on_` fields in diff and protocol logic

FeodorFitsner:
Updates diffing and protocol encoding logic to distinguish 'on_' fields that are not events using metadata. Adds integration tests for color scheme, new test cases for 'on_' fields, and updates ColorScheme to mark non-event 'on_' fields. This improves accuracy of change detection and serialization for controls with 'on_' prefixed fields.

## Summary by Sourcery

Distinguish non-event on_* fields in controls by using metadata and update diffing, protocol encoding, and attribute assignment logic accordingly, with new tests to validate behavior

Enhancements:
- Mark ColorScheme on_* fields as non-events via metadata to prevent incorrect event handling
- Restrict diff and protocol encoding to boolean-convert only on_* fields marked as events
- Simplify attribute setting logic by removing special-case conversion for on_* fields

Tests:
- Add unit tests for object diffing non-event on_* fields in frozen and in-place scenarios
- Add integration test for ColorScheme rendering with on_* color values

### DIFF
--- a/sdk/python/packages/flet-map/src/flet_map/marker_layer.py
+++ b/sdk/python/packages/flet-map/src/flet_map/marker_layer.py
@@ -28,7 +28,7 @@ class Marker(ft.Control):
     The coordinates of the marker.
 
     This will be the center of the marker,
-    if [`alignment`][(c).] is [`Alignment.CENTER`][flet.Alignment.].
+    if [`alignment`][(c).] is [`Alignment.CENTER`][flet.].
     """
 
     rotate: Optional[bool] = None
--- a/sdk/python/packages/flet/mkdocs.yml
+++ b/sdk/python/packages/flet/mkdocs.yml
@@ -143,7 +143,7 @@ plugins:
             preload_modules: [ flet, flet_ads ]
             filters:
               - "!^_" # Exclude private members starting with only one underscore
-              - "!(init|before_update|build|will_unmount|did_mount)"
+              - "!^(init|before_update|build|will_unmount|did_mount)$"
             extensions:
               - griffe_modernized_annotations
               - griffe_warnings_deprecated


sdk/python/packages/flet/integration_tests/controls/theme/test_color_scheme.py
```diff
@@ -0,0 +1,74 @@
...
+            ft.Text("Container surface", color=ft.Colors.ON_SURFACE_VARIANT, size=20),
+            width=200,
+            height=100,
+            bgcolor=ft.Colors.SURFACE,
+        ),
+    )
+    await flet_app.tester.pump_and_settle()
+
+    flet_app.assert_screenshot(
+        "cupertino_action_sheet_basic",
```
ndonkoHenri: Typo?

sdk/python/packages/flet/integration_tests/controls/theme/golden/macos/color_scheme/cupertino_action_sheet_basic.png
ndonkoHenri, line 1: File name looks wrong.

sdk/python/packages/flet/src/flet/controls/object_patch.py
```diff
@@ -963,7 +970,9 @@ def _compare_dataclasses(self, parent, path, src, dst, frozen):
                 if "skip" not in field.metadata:
                     old = getattr(src, field.name)
                     new = getattr(dst, field.name)
-                    if field.name.startswith("on_"):
+                    if field.name.startswith("on_") and (
+                        "event" not in field.metadata or field.metadata["event"]
+                    ):
```
ndonkoHenri, line 974: Can the if-expressions (here and in `protocol.py`) be improved to something like: 
`if field_name.startswith("on_") and fields[field_name].metadata.get("event", True):` ?

packages/flet/lib/src/controls/list_view.dart
```diff
@@ -79,14 +79,31 @@ class _ListViewControlState extends State<ListViewControl> {
...
-                itemExtent: itemExtent,
-                prototypeItem: prototypeItem,
-                children: controls
-                    .map((item) => ControlWidget(
-                          key: ValueKey(item.getKey("key")?.value ?? item.id),
-                          control: item,
-                        ))
-                    .toList(),
+                itemExtent: spacing > 0 ? null : itemExtent,
+                prototypeItem: spacing > 0 ? null : prototypeItem,
```
ndonkoHenri, line 84: Can this be mentioned in the python documentation of these props?
Kind of:
```
Note:
    Has effect only when spacing is ...
```

This makes me actually question: spacing is usually a value greater than or equal to 0; so in what case are these props set? when spacing is = 0 only?

packages/flet/lib/src/controls/view.dart
```diff
@@ -76,8 +76,6 @@ class _ViewControlState extends State<ViewControl> {
         if (_popCompleter != null && !_popCompleter!.isCompleted) {
           _popCompleter?.complete(args["should_pop"]);
         }
-      default:
-        throw Exception("Unknown View method: $name");
```
ndonkoHenri, line 80: why remove?

FeodorFitsner: View is "inherited" from  scrollable, so when `view.scroll_to()` is called View does not crash but rather passes the call down to a scrollable.

packages/flet/lib/src/controls/list_view.dart
FeodorFitsner: If spacing is not zero setting item extent or prototype does not have any sense as spacing adds width/height to item extent.

I would think we can try wrapping prototype item into row/column and add a spacer, but then what would be the hight/width of the last element?..

FeodorFitsner: I don't think item extent and prototype helps a lot as it's "non-build-on-demand" mode which is not default and should be used for building relatively small lists. For large lists built on-demand we use `ListView.separated()` and `ListView.builder()`.

sdk/python/packages/flet/src/flet/controls/core/list_view.py
```diff
@@ -54,6 +54,9 @@ class ListView(LayoutControl, ScrollableControl, AdaptiveControl):
     """
     A fixed height or width (when [`horizontal`][(c).] is `True`)
     of an item to optimize rendering.
+
+    Note: This property has effect only when [`build_controls_on_demand`][(c).]
+        is `True` or [`spacing`][(c).] is `0`.
```
ndonkoHenri, line 60: image

"Note:" must be alone one one line.
```
Note:
    text here...
```

PR #5774, opened 2025-11-04
feat(flet-video): enter/exit `Video` fullscreen programmatically

### LINKED ISSUE #3732: Programmatically enter fullscreen mode of Video control
### Duplicate Check

- [X] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

Video Control does not have a method to programmatically enter fullscreen mode. So, for example, we can bind some buttons to play or pause a video, but we cannot do the same for fullscreen mode.

### Suggest a solution

_No response_

### Screenshots

_No response_

### Additional details

_No response_

ndonkoHenri:
Fix #3732

## Summary by Sourcery

Enable programmatic fullscreen support for the Video control by exposing a fullscreen property, handling native fullscreen transitions in Flutter, and updating the Python API and example to demonstrate the new feature while marking the next package versions as Unreleased.

New Features:
- Add fullscreen boolean property to Video control for programmatic fullscreen toggling
- Implement enter/exit fullscreen handlers in Flutter VideoControl with a GlobalKey and native fullscreen transitions
- Expose programmatic fullscreen in Python Video API and update CHANGELOGs across packages

Enhancements:
- Revamp Python example to include an 'Enter Fullscreen' button and wrap content in SafeArea
- Refine debug logging and restructure example layout with Column controls

### DIFF
--- a/sdk/python/packages/flet-video/CHANGELOG.md
+++ b/sdk/python/packages/flet-video/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [0.2.0] - Unrealeased
+## [0.2.0] - Unreleased
 
 ### Added
 


sdk/python/examples/controls/video/example_1.py
```diff
@@ -1,141 +1,139 @@
...
-        print("Video.play_or_pause()")
 
-    async def handle_play(e):
+    async def handle_play(e: ft.Event[ft.Button]):
         await video.play()
-        print("Video.play()")
 
-    async def handle_stop(e):
+    async def handle_stop(e: ft.Event[ft.Button]):
         await video.stop()
```
FeodorFitsner, line 47: ...or you can just do a handler without argument if you don't need anything from `e`:

```py
async def handle_stop():
    await video.stop()
```

sdk/python/packages/flet-video/src/flutter/flet_video/lib/src/video.dart
```diff
@@ -31,6 +32,24 @@ class _VideoControlState extends State<VideoControl> with FletStoreMixin {
   late final controller =
       VideoController(player, configuration: videoControllerConfiguration);
 
+  Future<void> _handleEnterFullscreen() async {
+    widget.control.updateProperties({"_fullscreen": true}, python: false);
```
FeodorFitsner, line 36: We have to re-think if we need these `python: false` at all.
I noticed in web mode if you do refresh that `_fullscreen` state, which is client-only, is lost. Right?

sdk/python/examples/controls/video/example_1.py
ndonkoHenri: I prefer type hinting, for educational purposes. 🙂

PR #5817, opened 2025-11-15
refactor!: consolidate `src_base64`, `src`, and `src_bytes` into unified/intelligent `src` prop

### LINKED ISSUE #5470: v1: Harmonize `src` attribute across controls
Instead of `src`, `src_base64` and `src_bytes` have a single `src` property that can carry on URL, path, base64 encoded value and array of bytes.

ndonkoHenri:
Fix #5470

## Example:
```python
import base64

import flet as ft
import flet_lottie as ftl

image_b64 = "iVBORw0KGgoAAAANSUhEUgAAABkAAAAgCAYAAADnnNMGAAAACXBIWXMAAAORAAADkQFnq8zdAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAA6dJREFUSImllltoHFUYx3/fzOzm0lt23ZrQ1AQbtBehNpvQohgkBYVo410RwQctNE3Sh0IfiiBoIAjqi6TYrKnFy4O3oiiRavDJFi3mXomIBmOxNZe63ay52GR3Zj4f2sTEzmx3m//TYf7/c35zvgPnO6KqrESXqpq3muocAikv6m+/zytj3ejik1VN21G31YA9CgJ6xC+bMyQZPVCuarciPAMYC99V6Vw5pLbFSibHmlVoRVj9P3cmPBM8tSJI/M6mzabpfoAQ9fIF7WK4bd5vvuFnLGgy2vi0abg94A0AcJGvMq3hDxGRyar9r4F+iLAm0yIiRk8m37tctS1WsrIhhrI30+Srmg+J87OXUf3lWGS1q89dC6ltsSanxk4Aj2QBABii96300g87P/rtlrWr8l+vyDMfdlXSyyEikqxsiOUAQJCBhfHdXRfCq1LSsSlcWG+KBAGStvvrMkgiuv8lUc2mREukPwLUfHG+uTQv8Eown7VL3XlbBxYhf1c17hbVF3MDwA9bts280TnaU1YYqPby07aeFlUlHt27wSQ4CLo+F8AvoTCvHmyKF+ZbEb/M77P2LgvAwmrTHAHflN3KZxVbMC2jMFNOpgPnrMSOhvvFkMezXdwV4ePbtvHtxnJAMQ0j4JtVnO+eLb5oiSlt5HDbv7t1O90lpYCCCKbhfzW5kAIwUAazR0BlfII8Ow0I6uoVmI9MyAMwbMs8CExmDbk4zgu931MyO4OI4KrYflkRjOoTI+uM9d1vjotwKPu9QMk/sxzuO8POiVFcdZ1M2YBVsMEAKOqLvaPIe7mACuw0z/80SMH58SMplxlfiDhVi7dw2pltRhjKBQTQdrSja2KKTfE551NHuaZ0QVPvWYQUn31/Vm2nDvgjF4grVJx6suSvrvrSJ/6cSW2Oz9mf264uNrB806xZ1k/CZ49dUKgDEtlCROX2hfHpx8pGuuo3PpqYulw8fjndOp1yhgtNKRevJ1FyR2Ola+jXAjdnwTkZ6o896GdWdxDw7IxFg+0DpmXchTKSBWQnIuJn9u4j7dt+13UfHXEkXQOcuQ4kMhVtqsgUyPiQiPQfHw1NB2sRjmXKuTg1NwwBYLhtPtQX26eqTwGXPDOqvmcC4Hnwfrrad94GrVsOYTqUTkQY+iTlNe/6O1miSP/x0VB/+wMIDwHn/vtV1iQC4Xv95uUEWVCoL9Y5Z+gdovoyMHUFJHv88jmVy0vTuw7cZNv2YaA61Bfb7ZX5F8SaUv2xwZevAAAAAElFTkSuQmCC"

with open("path/to/lottie.json", "rb") as f:
    lottie_bytes = f.read()


def main(page: ft.Page):
    page.add(
        ft.Row(
            controls=[
                # ftl.Lottie(src="lottie.json"),
                ftl.Lottie(
                    src="https://raw.githubusercontent.com/xvrh/lottie-flutter/master/example/assets/Mobilo/A.json",
                ),
                ftl.Lottie(src=lottie_bytes),
                ftl.Lottie(src=base64.b64encode(lottie_bytes).decode("utf-8")),
            ]
        ),
        ft.Row(
            controls=[
                # ft.Image(src="image.png", width=100, height=100),
                ft.Image(
                    src="https://flet.dev/img/logo.svg",
                    width=100,
                    height=100,
                ),
                ft.Image(
                    src=b64,
                    width=100,
                    height=100,
                ),
                ft.Image(
                    src=base64.b64decode(image_b64),
                    width=100,
                    height=100,
                ),
            ]
        ),
        ft.Row(
            controls=[
                ft.CircleAvatar(
                    foreground_image_src=image_b64,
                    width=100,
                    height=100,
                ),
                ft.CircleAvatar(
                    foreground_image_src=base64.b64decode(image_b64),
                    width=100,
                    height=100,
                ),
            ]
        ),
    )


if __name__ == "__main__":
    ft.run(main, assets_dir="assets")
```


packages/flet/lib/src/utils/images.dart
```diff
@@ -105,31 +121,260 @@ bool isUrlOrPath(String value) {
...
+      src = src.trim();
+      if (src.isEmpty) {
+        return const ResolvedAssetSource();
+      }
+
+      // Base64
+      if (src.isBase64) {
+        final bytes = base64Decode(src.stripBase64DataHeader());
+        return ResolvedAssetSource(bytes: bytes);
+      }
```
FeodorFitsner: Checking for bytes list first is right way to go.

Now string branch. I'm thinking - could it be the way to determine if a string is URL or path first and only then fall back to base64? Or if it's URL then it's not a big performance hit to try converting small string into a base64 array?

ndonkoHenri: > I'm thinking - could it be the way to determine if a string is URL or path first and only then fall back to base64?

Nice idea actually. What do you think of this util, which i could reuse:
https://github.com/flet-dev/flet/blob/d013f4f5d32d9b62b4591b385285323c735a758e/packages/flet/lib/src/utils/images.dart#L108-L122

Only fear is whether is fully general (i.e. covers all possible edge cases). Or we can adjust with time, when users report issues?

FeodorFitsner: Checking URL like that is fine I guess - I think it could cover 99% of cases.
Files though - that could be tricky. For example, `abc` is a file/path.

OK, I think the chain could be as following:
1. Check if it's a list of bytes.
2. Check if it satisfies URL regexp.
3. Check if it contains `.` - it's a path as base64 doesn't allow `.`.
4. Quick check using [this regexp](https://stackoverflow.com/a/8571649/1435891) if a string **could be** a base64 string, then, if true, try converting it to bytes (which I'm pretty sure fails fast as well) and if it fails consider string as path.

ndonkoHenri: Concerning the base64 regex you linked:
image
image

I pushed a [commit](https://github.com/flet-dev/flet/pull/5817/commits/5b3a046f58c013ed7f7e93905334193bfecb3fbb), let me have your thoughts.

packages/flet/lib/src/utils/images.dart
```diff
@@ -342,20 +332,26 @@ class ResolvedAssetSource {
...
+      if (isUrl(src)) return ResolvedAssetSource(uri: src);
+
+      // asset path
+      if (src.contains(".")) return ResolvedAssetSource(uri: src);
 
       // Base64
       if (src.isBase64) {
-        final bytes = base64Decode(src.stripBase64DataHeader());
-        return ResolvedAssetSource(bytes: bytes);
+        final srcAsBytes = base64Decode(src.stripBase64DataHeader());
```
FeodorFitsner: I think it's fine to check base64 validity by trying to convert from base64.
But it should be made in one pass. Here, you are decoding first time to check if it's base64 and, if it's true, you are decoding it second time.

It should be one pass, kind of decode from base64 in a try-catch.

packages/flet/lib/src/controls/canvas.dart
```diff
@@ -541,16 +539,19 @@ Future<void> loadCanvasImage(Control shape) async {
   final completer = Completer();
   shape.properties["_loading"] = completer;
 
-  final src = shape.getString("src");
-  final srcBytes = shape.get("src_bytes") as Uint8List?;
+  final resolved = ResolvedAssetSource.from(shape.get("src"));
```
FeodorFitsner: I'd rename `resolved` to `src` and maybe wrapped `ResolvedAssetSource.from` to `getSrc`?

```dart
var src = shape.getSrc("src")
...
if (src.error != null) ...

if (src.hasUri) ...
```

packages/flet/lib/src/controls/cupertino_switch.dart
FeodorFitsner, line 92: should be `get` to support different image sources?

FeodorFitsner: also, in avatar control it's `"foreground_image_src"`. Shall we use `_src` everywhere or not?

FeodorFitsner (approved):

FeodorFitsner (approved):

PR #5901, opened 2025-12-13
fix: Ensure consistent `TextStyle` inheritance during button state transitions

### LINKED ISSUE #5278: CrossAxisAlignment.STRETCH not working with wrapped layout controls
This affects at least `ResponsiveRow` and `Row` with `wrap=True`. Whenever their child controls span to second row, the `vertical_alignment+CrossAxisAlignment.STRETCH` property stops working. 

The underlying problem appears to be that whenever their controls are wrapped, the controls change to `Wrap` flutter widget.  The `vertical_alignment` for `Wrap` uses [WrapCrossAlignment](https://api.flutter.dev/flutter/rendering/WrapCrossAlignment.html), which does not support `stretch` (or `baseline`) value.

Since this is something missing from Flutter, there probably is no easy fix. Perhaps the Flet controls could be rewritten to not use `Wrap`, but that may not be a simple task.

In the meanwhile, I think it would be appropriate to document this and/or even print a warning whenever `CrossAxisAlignment.STRETCH` or `CrossAxisAlignment.BASELINE` is used with a wrapping control.

### LINKED ISSUE #5620: V1: When a TextStyle is provided through Theme.button_theme.style.text_style, the app throws a Flutter-side exception (only reproducible in debug mode)
### Duplicate Check

### Describe the bug

When a TextStyle is provided through Theme.button_theme.style.text_style, the app throws a Flutter-side exception (only reproducible in debug mode). The same text_style applied directly on a ButtonStyle for individual buttons does not crash.

Also, when using button_theme, there is a visible delay before the style is applied to the button, maybe 1 second. It doesn't happen when the style is applied directly in button.

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft


def main(page: ft.Page):
    page.title = "Basic elevated buttons"
    page.theme = ft.Theme(
        button_theme=ft.ButtonTheme(
            style=ft.ButtonStyle(
                text_style=ft.TextStyle(
                    size=20,
                    italic=True,
                    # bgcolor=ft.Colors.PURPLE,
                    # color=ft.Colors.ORANGE,
                ),
            ),
        )
    )
    page.add(
        ft.Button(
            "Button",
        ),
    )


ft.run(main)
```

</details>

### To reproduce

1. run repro code `uv run flet run -w -p 8550 playground/button_theme_example.py`
2. `fvm flutter run macos -d`

### Expected behavior

_No response_

### Screenshots / Videos

<details open>
<summary>Captures</summary>

[Upload media here]

</details>

### Operating System

macOS

### Operating system details

macOS

### Flet version

0.69.0

### Regression

No, it isn't

### Suggestions

_No response_

### Logs

```console
[Paste your logs here]
```

</details>

### Additional details

_No response_

ndonkoHenri:
Fix #5620
Fix #5278

## Summary by Sourcery

Improve button text style handling and polish layout, utilities, and documentation across Flutter and Python SDK components.

Bug Fixes:
- Ensure button text styles resolve with non-inheriting TextStyle across widget states to avoid animation and lerp assertion issues.

Enhancements:
- Refine Row and ResponsiveRow layout behavior wiring to use higher-level alignment helpers and respect intrinsic height flags directly from control properties.
- Refactor text utilities by extracting font variation and text decoration parsing into dedicated helpers for reuse and clarity.
- Tighten pip helper error reporting and exception handling when ensuring Flet-related Python packages are installed.
- Clarify Lottie, Row, and ResponsiveRow Python API docstrings with concrete examples, alignment constraints, and current feature limitations.
- Adjust button content construction to reuse prebuilt content widgets instead of rebuilding inline.

Build:
- Extend Python isort configuration to treat additional Flet packages (desktop and web) as first-party modules.


sdk/python/packages/flet/src/flet/messaging/protocol.py
```diff
@@ -73,8 +73,13 @@ def encode_object_for_msgpack(obj):
...
         elif isinstance(obj, (datetime.datetime, datetime.date)):
-            if isinstance(obj, datetime.datetime) and obj.tzinfo is None:
-                obj = obj.astimezone()
+            if isinstance(obj, datetime.datetime):
+                if obj.tzinfo is None:
+                    # interprete naive datetimes as UTC
+                    obj = obj.replace(tzinfo=datetime.timezone.utc)
+
+                # Normalize to UTC
+                obj = obj.astimezone(datetime.timezone.utc)
```
FeodorFitsner: If I'm not mistaken Flutter DateTimePicker (or other widgets) expect (and outputs) local dates, not UTC.

Converting to UTC date on Python side could give wrong results on UI.

For example, if I specify `datetime(2025, 3, 10)` in Python and you convert it to UTC then on a device in -07:00 timezone it will show March 9 in the picker, instead of expected March 10.

When a developer writes a naive date like `2025-10-03` in Python, most people mentally model it as a local calendar date, not UTC. Usually, UTC must be explicit.

On the other hand, using of UTC for passing datetime/time over a wire (like in Flet between Python and Flutter) is a good thing. However, on Flutter (or JavaScript in web) side you have to convert the date back to local, either on protocol level or in DateTimePicker using `.toLocal()`. Selected date in DatePicker should be converted back to UTC with `.toUtc()`.

So, algorithm could be:
- "naive" datetime we convert to a local datetime then into UTC.
- local datetime is converted to UTC.
- UTC datetime goes into the wire "as is".

Does it make sense?

FeodorFitsner: ...why passing UTC over the wire is the right solution? When you have both Python and Flutter on the same machine (mobile, desktop apps) it's fine to use local, but if you run Flet app on the web then client browsers are in time zones different from server: client receives UTC date and converts it to its local date.

ndonkoHenri: I reverted the timezone changes in this PR. Will be handled in another one.

FeodorFitsner: Good, created an issue: https://github.com/flet-dev/flet/issues/5923

PR #6013, opened 2026-01-15
Fix Android platform check to exclude web

FeodorFitsner:
Updated the platform check to ensure Android-specific code does not run on web by adding a '!kIsWeb' condition. This prevents potential issues when running on web platforms that report as Android.

## Summary by Sourcery

Bug Fixes:
- Prevent Android-only device info logic from running when the app is executing in a web environment.

### DIFF
--- a/packages/flet/lib/src/flet_backend.dart
+++ b/packages/flet/lib/src/flet_backend.dart
@@ -253,7 +253,7 @@ class FletBackend extends ChangeNotifier {
         await pageSizeUpdated.future;
         debugPrint("Registering web client with route: $newRoute");
         String platform = defaultTargetPlatform.name.toLowerCase();
-        if (platform == "android") {
+        if (platform == "android" && !kIsWeb) {
           try {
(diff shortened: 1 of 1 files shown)


packages/flet/lib/src/flet_backend.dart
```diff
@@ -253,7 +253,7 @@ class FletBackend extends ChangeNotifier {
         await pageSizeUpdated.future;
         debugPrint("Registering web client with route: $newRoute");
         String platform = defaultTargetPlatform.name.toLowerCase();
-        if (platform == "android") {
+        if (platform == "android" && !kIsWeb) {
```
ndonkoHenri, line 256: You can use this util, i guess:

https://github.com/flet-dev/flet/blob/16952ae0b8e14223c622e9bcac1a3c0a624b687a/packages/flet/lib/src/utils/platform.dart#L41-L44

FeodorFitsner: I thought so too, but I need `platform` value after that block.

PR #6035, opened 2026-01-19
Ubuntu 20.04-24.04

FeodorFitsner:
## Summary by Sourcery

Expand Linux client build and packaging to support multiple Debian and Ubuntu distributions and align desktop client resolution with distribution-specific packaging.

New Features:
- Add per-distribution Linux build matrix entries for Debian 10/12 and Ubuntu 20.04/22.04/24.04 across amd64 and arm64 architectures.
- Introduce distribution-aware storage paths for unpacked desktop clients based on the installed flet-desktop package variant.

Enhancements:
- Update Linux build artifacts and wheel tagging to be distro- and manylinux-tag aware, including renamed tarball and artifact outputs.
- Gate Rive client inclusion behind an explicit configuration flag rather than architecture alone.
- Improve git safety configuration within CI containers and normalize version patching scripts to only touch uninitialized version fields.
- Require desktop client archives to be present in the package installation instead of downloading them at runtime.

Build:
- Run Linux builds inside distro-specific container images with additional required system packages and more robust apt handling, including support for archived Debian 10 repositories.

CI:
- Refine Linux CI job definitions with distribution identifiers, manylinux tags, and updated artifact naming for clearer matrix runs and outputs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,31 +131,15 @@ jobs:
   # Build Flet Client for Windows
   # =============================
   build_windows:
-    name: Build Flet Client for Windows ${{ matrix.title }}
-    runs-on: ${{ matrix.runner }}
+    name: Build Flet Client for Windows
+    runs-on: windows-latest
     needs:
       - python_tests
       - build_flet_package
-    strategy:
-      matrix:
-        include:
-          - arch: x64
-            build_arch: x64
-            runner: windows-latest
-            wheel_tag: win_amd64
-            title: X64
-          - arch: arm64
-            build_arch: arm64
-            runner: windows-11-arm
-            wheel_tag: win_arm64
-            title: ARM64
     env:
       PKG_VER: ${{ needs.build_flet_package.outputs.PKG_VER }}
       BUILD_NUM: ${{ needs.build_flet_package.outputs.BUILD_NUM }}
       PYPI_VER: ${{ needs.build_flet_package.outputs.PYPI_VER }}
-      ARCH: ${{ matrix.arch }}
-      BUILD_ARCH: ${{ matrix.build_arch }}
-      WHEEL_TAG: ${{ matrix.wheel_tag }}
     steps:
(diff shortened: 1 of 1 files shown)


sdk/python/packages/flet-desktop/src/flet_desktop/__init__.py
```diff
@@ -204,16 +209,3 @@ def __locate_and_unpack_flet_view(page_url, assets_dir, hidden):
...
-    ver = flet_desktop.version.version
-    if not ver:
-        import flet.version
-
-        ver = flet.version.flet_version
-    temp_arch = Path(tempfile.gettempdir()).joinpath(file_name)
-    flet_url = f"https://github.com/flet-dev/flet/releases/download/v{ver}/{file_name}"
-    logger.info(f"Downloading Flet v{ver} from {flet_url} to {temp_arch}")
-    urllib.request.urlretrieve(flet_url, temp_arch)
-    return str(temp_arch)
```
ndonkoHenri, line 219: Can you please explain the plan here? Like, we no longer download, whats new behaviour?

FeodorFitsner: Flet client must be (and it's like it's now) part of a wheel downloaded from pypi. Downloads from GitHub have not been working since I don't remember when. It's leftovers cleanup.

The only case when that download could have triggered is installing from a source archive (.tar.gz on pypi) if the pre-built wheel was not found. But even in this case it could download non-supported binary from GitHub. Say, you are on RaspberryPI or it's Alpine Linux (muslinux tag) which is not supported by Flutter at all. Flet will try to download what from GitHub, "generic" linux client? It won't work, right? Or it's an old macOS - the same - premacOS client pre-built for newer macOS won't work.

FeodorFitsner: Other words, if it was unable to find a wheel for your platform then it's probably not supported.

PR #6043, opened 2026-01-20
Lazy loading of icons, theme for faster app startup

FeodorFitsner:


sdk/python/packages/flet/src/flet/controls/material/dropdown.py
```diff
@@ -196,7 +195,7 @@ class Dropdown(LayoutControl):
     Configures how the text input should be capitalized.
     """
 
-    trailing_icon: IconDataOrControl = Icons.ARROW_DROP_DOWN
+    trailing_icon: Optional[IconDataOrControl] = None
     """
     An icon to display at the end of the text field.
     """
```
ndonkoHenri, line 201: Can you mention the default in the docstring, with cross-ref [flet.] ?

sdk/python/packages/flet/src/flet/controls/material/dropdown.py
```diff
@@ -209,7 +208,7 @@ class Dropdown(LayoutControl):
     be aligned with the text in the text field.
     """
 
-    selected_trailing_icon: IconDataOrControl = Icons.ARROW_DROP_UP
+    selected_trailing_icon: Optional[IconDataOrControl] = None
     """
     An optional icon at the end of the text field to indicate that the text field is
     pressed.
```
ndonkoHenri, line 214: Can you mention the default in the docstring, with cross-ref [flet.] ?

ndonkoHenri:
New failing test btw: https://github.com/flet-dev/flet/actions/runs/21192744715/job/60962467504?pr=6043

FeodorFitsner:
> lgtm
> 
> So, the below changes are always to be applied, right?
> 
> * Theme -> "Theme"
> * icons import in before_update()

Looks like, yeah. Both are heavy things to be imported without necessity.
I'm actually thinking if that's even possible to do everything lazy loaded? :)

sdk/python/packages/flet/src/flet/controls/material/dropdown.py
FeodorFitsner: It was not needed to explicitly set icons at all. They are nullable on Flutter side and if not specified, arrows are used.

yst-one:
> > lgtm  
> > So, the below changes are always to be applied, right?那么，以下更改总是要应用，对吗？
> > 
> > * Theme -> "Theme"
> > * icons import in before_update()icons 在 before_update() 中导入
> 
> Looks like, yeah. Both are heavy things to be imported without necessity.看起来是的。这两项在没有必要的情况下导入都很重。 I'm actually thinking if that's even possible to do everything lazy loaded? :)我其实在想，是否真的可以做到所有内容都懒加载呢？ :)
It's so needed.

Issue #6047, opened 2026-01-21
bug: ButtonStyle in TextButton has no effect


lastdayafter:
### Duplicate Check

- [x] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

Applying ButtonStyle to TextButton has no effect.

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft


def main(page: ft.Page):
    page.window.width = 400
    page.window.height = 400
    page.vertical_alignment = ft.MainAxisAlignment.CENTER
    page.horizontal_alignment = ft.CrossAxisAlignment.CENTER

    page.add(
        ft.TextButton(
            content='test_button_0',
            style=ft.ButtonStyle(
                bgcolor=ft.Colors.WHITE,
                color=ft.Colors.BLACK,
                shape=ft.RoundedRectangleBorder(0),
            )
        )
    )


if __name__ == '__main__':
    ft.run(main)
```

</details>

### To reproduce

Run the code

### Screenshots / Videos

<details open>
<summary>Captures</summary>

<img width="513" height="514" alt="Image" src="https://github.com/user-attachments/assets/f1338d0e-3419-47df-bd7c-045473d3c8fb" />

</details>

### Operating System

macOS

### Operating system details

26.2

### Flet version

0.80.1

### Regression

I'm not sure / I don't know

ndonkoHenri:
Issue has been fixed in Flet v0.80.4.

Issue #6044, opened 2026-01-21
bug: [Flet 0.80.1] How can i expand a dropdown to all horizontal available space ?


NiTRoeSE:
### Duplicate Check

- [x] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

Hello,

in flet 0.28.3 my dropdowns are correct expanded to all available space using `expand=True`.
Unfortunately this doesnt work anymore in flet 0.80.1.

Its possible to set a width, but at this point i cant access `page.width` in my app, the other problem with setting a specific width is that even `menu_width` key is set the menu goes over the whole page.

I tried also to place the dropdown in a `ft.ResponsiveRow` and gave `col={"sm":12}` but that also didnt work.
I think `expand=True` should work and its maybe a bug that it do nothing on dropdown, i dont know.. 

Is there any way to let a dropdown use all available space in a page / column / row / container without using the `width` key ?

Thanks in advanced

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft

def main(page: ft.Page):
    page.spacing = 20
    page.padding = 20

    severity_dropdown =  ft.Dropdown(
                    
                        label="Severity",
                        leading_icon=ft.Icons.WARNING_ROUNDED, 
                        border_color=ft.Colors.ON_TERTIARY_CONTAINER,
                        fill_color=ft.Colors.ON_TERTIARY,
                        filled=True,
                        menu_height=400,
                        menu_width=400,
                        options = [
                            ft.DropdownOption(
                                key="warn",
                                content=ft.Text(
                                    value="Warning (Orange)",
                                    color=ft.Colors.AMBER_400
                                ),
                                
                            ),
                            ft.DropdownOption(
                                key="success",
                                content=ft.Text(
                                    value="Success (Green)",
                                    color=ft.Colors.GREEN_400
                                ),
                                
                            ),
                            ft.DropdownOption(
                                key="info",
                                text="Information (White)"
                            ),
                            ft.DropdownOption(
                                key="danger",
                                content=ft.Text(
                                    value="Danger (Red)",
                                    color=ft.Colors.RED_400
                                ),
                               
                            ),

                        ],
                        expand=True,
                        hover_color=ft.Colors.TRANSPARENT,
                        
                        
        )

    page.add(severity_dropdown)

ft.run(main)

```

</details>

### To reproduce

Run code with 0.28.3 --> Works
Run code with 0.80.1 --> Didnt work

[...]

PR #6048, opened 2026-01-21
fix: Enable TextButton style and full-width Dropdown

7576457:
## Description

Made TextButton style actually work and allowed Dropdown to expand fully

issues:
* #6047
* #6044

## Type of change

## Checklist

- [x] I signed the CLA.
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing tests pass locally with my changes
- [x] I have made corresponding changes to the [documentation](https://github.com/flet-dev/website) (if applicable)

## Screenshots 

image
image

## Summary by Sourcery

Fix button styling and dropdown layout behavior to honor configured styles and horizontal expansion.

Bug Fixes:
- Make button controls correctly apply the configured style from control properties.
- Allow dropdown controls to fully expand horizontally when the expand property is set.

### DIFF
--- a/packages/flet/lib/src/controls/button.dart
+++ b/packages/flet/lib/src/controls/button.dart
@@ -98,7 +98,7 @@ class _ButtonControlState extends State<ButtonControl> with FletStoreMixin {
     var theme = Theme.of(context);
 
     var style = parseButtonStyle(
-      widget.control.internals?["style"],
+      widget.control.internals?["style"] ?? widget.control.get("style"),
       theme,
       defaultForegroundColor:
           widget.control.getColor("color", context, theme.colorScheme.primary)!,
--- a/sdk/python/packages/flet/src/flet/controls/material/text_button.py
+++ b/sdk/python/packages/flet/src/flet/controls/material/text_button.py
@@ -101,10 +101,5 @@ class TextButton(LayoutControl, AdaptiveControl):
     Called when this button has lost focus.
     """
 
-    def before_update(self):
-        super().before_update()
-        if self.style is not None:
-            self._internals["style"] = self.style.copy()
-
     async def focus(self):
         await self._invoke_method("focus")


packages/flet/lib/src/controls/button.dart
```diff
@@ -98,7 +98,7 @@ class _ButtonControlState extends State<ButtonControl> with FletStoreMixin {
     var theme = Theme.of(context);
 
     var style = parseButtonStyle(
-      widget.control.internals?["style"],
+      widget.control.properties["style"],
       theme,
```
FeodorFitsner, line 102: I see we use the same `button.dart` logic for all buttons, but on Python side we have inconsistency. "Regular" Button [puts style into `_internals`](https://github.com/flet-dev/flet/blob/main/sdk/python/packages/flet/src/flet/controls/material/button.py#L143-L147), but TextButton still uses `style` property. We need to have similar Python logic for all buttons to fix that, otherwise your change fixes TextButton, but break Button.

ndonkoHenri: Thats right. We could also do something like below if we wanna keep current logic (avoid using python `internals` prop in `TextButton` if not really useful): 
```dart
widget.control.internals?["style"] ?? widget.control.get("style")
```

packages/flet/lib/src/controls/dropdown.dart
```diff
@@ -291,6 +292,7 @@ class _DropdownControlState extends State<DropdownControl> {
       hintText: widget.control.getString("hint_text"),
       helperText: widget.control.getString("helper_text"),
       menuStyle: menuStyle,
+      expandedInsets: widget.control.getExpand("expand") == 1 ? EdgeInsets.zero : null,
```
ndonkoHenri: Is this property worth exposing on Python end?

also, `==1` here seems limited in some way? cause what if i set `expand=3` in my app, and give another horizontally placed item `expand=1`? We would in that case still expect the DD to expand fully horizontally, right?

FeodorFitsner: I thought the same. Maybe new property then? ;)

packages/flet/lib/src/controls/button.dart
7576457: In that case, I would bring `ft.TextButton` to the same behavior as other buttons, for example, in the `ft.TextButton` component itself:

```python
def before_update(self):
    super().before_update()
    if self.style is not None:
        self._internals["style"] = self.style.copy()
```

Also, I'm curious to know why the additional `_internals` parameter is needed? Why can't the `style` parameter be kept in the same place where other parameters are in `properties`?

packages/flet/lib/src/controls/dropdown.dart
7576457: Hmm, I didn't think about that :)
In that case, I could suggest the following:
```dart
int? expand = widget.control.getExpand("expand");
EdgeInsets? expandedInsets = (expand ?? 0) > 0 
    ? EdgeInsets.all((expand ?? 0).toDouble()) 
    : null;
```

image

7576457: I'm not sure if this property needs to be exposed in Python. I had an idea about somehow using the real Expand widget instead of padding :). But at the moment, I can't figure out how that would work

packages/flet/lib/src/controls/button.dart
ndonkoHenri: https://github.com/flet-dev/flet/blob/80b78cdf95fe0db404c421fbf75abc5b7b4546b5/sdk/python/packages/flet/src/flet/controls/material/button.py#L137-L147

In this code, we merge some individual props (individualized for convenience/dev-x) and the `style`, so we can send it over as one resulting `style`. Inorder to have this resulting style in properties, we will obviously need to modify the user's `Button.style` prop, which we want to avoid. Hence we send it through the wire in another way: `_internals`.

I will suggest you instead do it the way below, as there is no real need to make use of `self._internals` - we can use `style` prop directly (in the fallback clause):

```dart
widget.control.internals?["style"] ?? widget.control.get("style")
```

packages/flet/lib/src/controls/dropdown.dart
```diff
@@ -271,6 +271,11 @@ class _DropdownControlState extends State<DropdownControl> {
 
     _focusNode.canRequestFocus = editable;
 
+    int? expand = widget.control.getExpand("expand");
+    EdgeInsets? expandedInsets = (expand ?? 0) > 0
+        ? EdgeInsets.all((expand ?? 0).toDouble())
+        : null;
```
ndonkoHenri: The Flutter docs say this:

```dart
  /// Defines the menu text field's width to be equal to its parent's width
  /// plus the horizontal width of the specified insets.
  ///
  /// If this property is null, the width of the text field will be determined
  /// by the width of menu items or [DropdownMenu.width]. If this property is not null,
  /// the text field's width will match the parent's width plus the specified insets.
  /// If the value of this property is [EdgeInsets.zero], the width of the text field will be the same
  /// as its parent's width.
  ///
  /// The [expandedInsets]' top and bottom are ignored, only its left and right
  /// properties are used.
  ///
  /// Defaults to null.
  final EdgeInsetsGeometry? expandedInsets;
```

So I would guess, if the `expand` is greater than `0`, we should set `expandedInsets` to `EdgeInsets.zero`?
```suggestion
    int expand = widget.control.getExpand("expand", 0)!;
    EdgeInsets? expandedInsets = expand > 0 ? EdgeInsets.zero : null;
```

7576457: Done

packages/flet/lib/src/controls/button.dart
7576457: Done

PR #6107, opened 2026-02-01
ci: add top-level workflow orchestrator

ndonkoHenri:
https://github.com/flet-dev/flet/actions/runs/21566901473

## Summary by Sourcery

Introduce a top-level CI workflow that orchestrates reusable build, release, and macOS integration test workflows based on detected file changes.

Build:
- Split the monolithic CI workflow into a lightweight orchestrator that conditionally calls a reusable build-and-release workflow and a macOS integration tests workflow using changed-file detection.
- Extract the existing build, packaging, publishing, and release jobs into a dedicated build-and-release reusable workflow with shared environment and concurrency configuration.
- Convert the macOS integration tests workflow into a callable workflow so it can be triggered from the main CI orchestrator instead of on every push/path change.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -3,6 +3,11 @@ name: Build & Release
 on:
   workflow_call:
   workflow_dispatch:
+#  push:
+#    branches:
+#      - '**'
+#    tags:
+#      - 'v[0-9]+.[0-9]+.[0-9]+'
 
 permissions:
   id-token: write


FeodorFitsner:
Isn't it easier to specify files filter on each workflow instead?

ndonkoHenri:
> Isn't it easier to specify files filter on each workflow instead?

Its possible, yh. But the main goal of this PR is to have one main ci dashboard.

image

I find it personally easier to navigate through the ci workflows for each ref/commit, as its all on one spot, especially as we now going to have 3 workflows all usually running at the same time. But let me know what you think.

PR #6183, opened 2026-02-18
feat: Add `ignore_up_down_keys` to `TextField` and `CupertinoTextField` to opt-out of default arrow-key caret jumps

### LINKED ISSUE #1423: How to Not jump to beginning of input text or the end when using arrow up or down in TextField?
When I type in TextField and pressed arrow key up or down, the cursor moves to the beginning or the end of the input text. How can I disable this behavior? The reason for this is that I want these two keys to perform some other actions such as select controls in a ListView.

I searched for a way to do it in Flutter, it can be done by using the RawKeyListener. Does Flet have similar functionality?

ndonkoHenri:
Fix #1423

## Summary by Sourcery

Introduce an option on text fields to suppress default up/down arrow key caret movement while centralizing key handling logic for submit behavior.

New Features:
- Add an ignore_up_down_keys configuration flag to TextField and CupertinoTextField to disable arrow-up and arrow-down caret jumps when enabled.

Enhancements:
- Refactor text field key handling into a shared handler that also manages Enter-based submit behavior for focused fields.

Documentation:
- Document the new ignore_up_down_keys option on the Python TextField control, explaining its effect on arrow key behavior.


sdk/python/packages/flet/src/flet/app.py
```diff
@@ -31,8 +31,12 @@
 
 logger = logging.getLogger("flet")
 
-if TYPE_CHECKING:
-    from flet.controls.page import Page
+AppCallable = Callable[[Page], Any]
```
FeodorFitsner:
```suggestion
AppCallable = Union[Callable[["Page"], Any], Callable[["Page"], Awaitable[Any]]]
```

PR #6267, opened 2026-03-06
feat: make `SharedPreferences` support `int`, `float`, `bool` and `list[str]`

### LINKED ISSUE #6304: feature: Support types other than str in shared_preferences
### Duplicate Check

Currently `SharedPreferences().set()` supports only string values.
This is because in the shared_preferences.dart we use the methods `prefs.setString` and `prefs.getString`. Trying to pass a variable with a different type fails.

The underlying shared_preferences library supports int/double/bool/str/List[str], albeit in different methods.

Using the additional types (Especially List[str]) can be convenient and I think should be supported.

### Suggest a solution

The API to these methods isn't obvious though, and I'd like to gather opinions on it.

set() api:

For almost all types, the `set()` method can be differentiated by the type of the `value` argument (If `value` is int then call prefs.setInt in flutter). The one problematic type for that is the String list. This is because Python lists are dynamic and any String list will have the type `list`. Do we want to assume that any list passed to this method is a list of strings and try pass it as such? If so, when do we ever check this? In Python or flutter?

get() api:

The get method is more complicated. A single `get` method isn't really possible because `shared_preferences` doesn't keep track of the types of the values that it saves.
Flet has no way of knowing which type is expected as the return type, so we can't automatically select the correct method.

From what I see, there are two ways to solve this:
- We can keep track of the type of preferences that are being set by setting an additional preference for every one of them, and then checking it when getting. While this could work, I don't think this is a great solution because it requires us as a library to manage shared preferences.
- We can have the same methods as in dart, `get_int/get_string/get_string_list()`. While very much not Pythonic, these methods will work. In such a case though we might want to differentiate the set methods by type too.

### Screenshots

_No response_

### Additional details

Currently the method signatures for SharedPreferences().get()/set() are not correct because the set method implies that `value` can be of type `Any` and the get method doesn't specify the return type is string. In either case, they should be corrected.

I am willing to implement this feature myself, I mainly want to talk about the API in order to (hopefully) reach a consensus.

ndonkoHenri:
Fix #6304 

## Summary by Sourcery

Extend SharedPreferences to support multiple primitive types and lists while improving platform build configuration handling and documentation.

New Features:
- Allow SharedPreferences to store and retrieve str, int, float, bool, and list[str] values across Python and Flutter layers.
- Document how to use subprocess to invoke external system commands from Flet apps, including Android-specific examples.

Bug Fixes:
- Make CLI boolean parsing for macOS entitlements and Android features/permissions case-insensitive to avoid unexpected failures.
- Correct documentation references for permissions resolution order and flet run docs paths.
- Fix ads package changelog by converting the Unreleased section into a tagged 0.82.0 release entry with proper issue and PR links.

Enhancements:
- Clarify and standardize CLI and TOML value formats for iOS, macOS, and Android build options and entitlements, with richer examples and template translations.
- Tighten SharedPreferences type validation and add a dedicated type alias for supported value types in the Python API.
- Improve type hints and navigation in Python docs and auth APIs, including OAuth token typing and cookbook link fixes.
- Refine Flet CLI build command help text and link to the full publishing guide.
- Update deep linking and publishing docs with additional platform references and concrete manifest/plist translation examples.
- Add Subprocess cookbook page to the documentation navigation.
- Clarify release-preparation skill instructions for changelog handling and author attribution.

Documentation:
- Expand platform-specific publishing docs (iOS, macOS, Android) with supported value forms, examples, and rendered template snippets.
- Add detailed deep linking examples for Android and iOS, showing how TOML config maps to platform manifests.
- Introduce a new cookbook recipe demonstrating safe subprocess usage, including launching Android activities from Flet apps.

Tests:
- Broaden SharedPreferences integration tests to cover all supported value types and rejection of unsupported types.

### DIFF
--- a/sdk/python/packages/flet-cli/src/flet_cli/commands/build_base.py
+++ b/sdk/python/packages/flet-cli/src/flet_cli/commands/build_base.py
@@ -252,8 +252,8 @@ class BaseBuildCommand(BaseFlutterCommand):
             action="extend",
             nargs="+",
             default=[],
-            help="Files and/or directories to exclude from the package "
-            ";can be used multiple times",
+            help="Files and/or directories to exclude from the package"
+            "; can be used multiple times",
         )
         parser.add_argument(
             "--clear-cache",
--- a/sdk/python/packages/flet/docs/publish/android.md
+++ b/sdk/python/packages/flet/docs/publish/android.md
@@ -278,7 +278,7 @@ flet build aab --android-signing-key-store-password value
 ```
 ///
 /// tab | `pyproject.toml`
-For security reasons, the key password is not read from `pyproject.toml` to
+For security reasons, the keystore password is not read from `pyproject.toml` to
 prevent accidental exposure in source control. See the other tabs for supported alternatives.
 ///
 /// tab | `.env`


sdk/python/packages/flet-cli/src/flet_cli/commands/build_base.py
```diff
@@ -251,7 +253,7 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
             nargs="+",
             default=[],
             help="Files and/or directories to exclude from the package "
-            "(can be used multiple times)",
+            ";can be used multiple times",
```
FeodorFitsner:
```suggestion
            "; can be used multiple times",
```

Plus remove ` ` (space) at the end of previous line.

sdk/python/packages/flet/docs/publish/android.md
```diff
@@ -278,7 +278,7 @@ flet build aab --android-signing-key-store-password value
 ```
 ///
 /// tab | `pyproject.toml`
-For security reasons, the keystore password is not read from `pyproject.toml` to
+For security reasons, the key password is not read from `pyproject.toml` to
```
FeodorFitsner:
```suggestion
For security reasons, the keystore password is not read from `pyproject.toml` to
```

PR #6298, opened 2026-03-11
Disable page auto-update if any .update() was called in the event handler

### LINKED ISSUE #6236: Disable page auto-update if any `.update()` was called in the event handler
1. This is replicating the bahavior of Flet 0.28.x. If at least one `.update()` was called inside event handler page auto-update will be skipped, for example:

```py
def on_click():
    status.value = "Status has changed"
    status.update()
```

2. Write docs article to explain different auto-update modes (auto-update off/on, relation to isolated controls, tips and tricks).

FeodorFitsner:
Close #6236

## Summary by Sourcery

Skip automatic page updates at the end of event handlers when an explicit update has already been called, and document and test the new auto-update behavior.

New Features:
- Introduce context tracking to detect whether an explicit .update() was called during the current handler and expose it via context helpers.

Bug Fixes:
- Prevent redundant double updates by skipping auto-update after handlers that already called .update().

Enhancements:
- Integrate the update-called tracking into session after_event handling and Page.update() to centralize update behavior control.

Documentation:
- Expand the getting-started guide with an explanation of automatic page updates and how to enable or disable auto-update at handler and global scope.

Tests:
- Add unit and integration tests to cover update-called flag behavior, after_event auto-update skipping, generator-style handlers, and Page.update() flag setting.

### DIFF
--- a/sdk/python/packages/flet/tests/test_skip_double_update.py
+++ b/sdk/python/packages/flet/tests/test_skip_double_update.py
@@ -15,7 +15,7 @@ def _fresh_update_behavior():
     """Ensure each test starts with a clean UpdateBehavior."""
     token = _update_behavior_context_var.set(UpdateBehavior())
     yield
-    _update_behavior_context_var.set(token.old_value)
+    _update_behavior_context_var.reset(token)
 
 
 def _make_session():


datumbox:
@FeodorFitsner just wanted to say that this looks like a MAJOR improvement on speed.

Previous versions of Flet would cause a significant lag on the Python-driven elements of the UI (as opposed to the actions handled natively by flutter; so any callback to python method). When I've debugged it, I've noticed that the queue was being filled with messages that couldn't be processed fast enough. Particularly problematic were callbacks related to scroll tracking which would fill the message queue. For this reason, I've avoided completely any scroll tracking mechanism as it would render the app unusable for seconds when the UI was "at rest" (fully loaded but the user would interact with it). This was mostly seen when there were a lot of controls loaded on a view (for example let's say you have a chat app that requires one control per message and a long list of messages).

With 0.83.0, it seems that the problem is resolved. I can't comment whether the scroll tracking events are causing issues still because I've removed them from the app, but I can 100% say that the app is fully responsive now with all major lag issues resolved.

I've have always suspected it was the automatic update causing the issues (as I'm handling myself the updates manually to batch them and do them optimally) but I thought there isn't anything to be done. I was wrong, this patch completely fixes the issue. I have previously tried to even turn off automatic updates but this didn't really fix issues (not sure why) and reverted it.

I thought to post this now and let you know. Great job!

My only suggestion is that perhaps it's worth adding an end-to-end dummy chat app that has 200-300 messages loaded as part of your tests. This will easier allow you to check performance issues when someone loads the app then clicks to any button with Python callback. The issue on previous versions were dead obvious; from 1-10sec delay depending on how many messages where queued. If you would like to explore this let me know and I can help.

PR #6381, opened 2026-04-05
fix: select platform-compatible icon format in find_platform_image

HG-ha:
## Summary

`find_platform_image` uses `glob.glob` to match icon files by base name, then blindly picks `images[0]`. On NTFS the alphabetical ordering causes `.icns` (macOS-only) to be selected over `.ico`/`.png` when building for Windows, making `flutter_launcher_icons` fail with `NoDecoderForImageFormatException`.

This PR introduces a per-platform preferred extension list so that the most appropriate format is selected for each build target:

- **Windows**: `.ico` > `.png` > others
- **macOS**: `.icns` > `.png` > others  
- **Linux/Web/Android/iOS**: `.png` > others

Unrecognized extensions still sort last, so existing setups with only `.png` files are completely unaffected.

## How to reproduce

1. Place `icon.icns`, `icon.ico`, and `icon.png` in `src/assets/`
2. Run `flet build windows -v`
3. Build fails:
   ```
   Found "icon" image at ...\icon.icns
   ✗ ERROR: NoDecoderForImageFormatException
   images/icon.icns
   ```

## Changes

- `sdk/python/packages/flet-cli/src/flet_cli/commands/build_base.py`:
  - Added `_PLATFORM_IMAGE_PREFERENCE` class variable mapping each target platform to a ranked list of preferred image extensions
  - Modified `find_platform_image` to sort glob results by platform preference before selecting the best match

## Summary by Sourcery

Bug Fixes:
- Fix selection of incompatible .icns icons on non-macOS targets by choosing the most appropriate image format per platform when multiple files share the same base name.


sdk/python/packages/flet-cli/src/flet_cli/commands/build_base.py
```diff
@@ -2272,7 +2286,13 @@ def find_platform_image(
         hash: HashStamp,
     ):
         """
-        Find first matching image file by base name and queue it for copy.
+        Find the best matching image file for the current target platform.
+
+        When multiple files share the same base name (e.g. ``icon.icns``,
```
FeodorFitsner: Thanks for the PR! Can you remove all double ticks "``" in comments please?

HG-ha: Done, removed all double backticks. Thanks for the review!

FeodorFitsner: :) Well, you removed all ticks, not replaced them with a single ticks. It's fine, it's a known issue with Claude/Codex. Put single ticks back please.

HG-ha: I now understand your meaning. The correction has been made.

FeodorFitsner:
Your PR adds `.svg` and `.bmp` to the matrix which are not supported by `flutter_launcher_icons` package. I'd trim these lists to only what `lutter_launcher_icons` actually decodes.

Also, there is an alternative, simpler approach: Instead of a full preference dict, you could just exclude known-incompatible extensions per platform (e.g. strip `.icns` on non-macOS, strip `.ico` on non-Windows) and keep the existing "pick first" behavior. That's a smaller change and addresses the actual bug without introducing a ranking system the project then has to maintain. What do you think?

HG-ha:
Good point — I'll simplify the approach: just filter out incompatible
extensions per platform (.icns on non-macOS, .ico on non-Windows)
before picking the first match. That keeps the change minimal.

PR #6402, opened 2026-04-10
feat: add hot-reload watcher to CrocoDocs CLI

ndonkoHenri:
## Summary by Sourcery

Add a CrocoDocs file-watching CLI workflow for hot-reload docs development.

New Features:
- Add a crocodocs watch CLI subcommand and supporting file-watching pipeline to regenerate docs artifacts on source changes and optionally run a child dev server.

Bug Fixes:
- Correct CrocoDocs manifest route generation so docs/index pages map to the base route rather than an /index suffix.

Enhancements:
- Improve CrocoDocs CLI UX with shared arguments, richer help/description text, and clearer watcher configuration flags.
- Refine CrocoDocs README to describe the watch workflow, watched paths, and configuration tables with clearer formatting.

Build:
- Update website and Taskfile docs commands so yarn start runs crocodocs watch alongside the Docusaurus dev server for hot-reload docs development.

Documentation:
- Improve CrocoDocs and website docs tables and copy to better explain keys, components, and generated artifacts.


ndonkoHenri:
@sourcery-ai review

FeodorFitsner:
Why I'd prefer `watchdog` over a hand-rolled polling watcher:

- It's the de-facto Python file-watch library — already a transitive dep in much of the Python ecosystem, not an exotic addition.
- **Event-based, zero CPU when idle.** The current polling re-stats every watched file every 750 ms. With the `examples_root` / `asset_mappings.examples` overlap, that's thousands of `stat()` syscalls per second on a laptop. Fine on SSD, noticeable on battery / HDD / network mounts.
- **Lower save→regen latency.** Drops from ~0.75–1.25 s (poll interval + debounce) to just the debounce window (~0.5 s). For hot-reload UX that's the difference between "fast" and "feels instant."
- **Cross-platform edge cases already handled:** case-insensitive filesystems, symlinks, and rename-as-delete+create. The current `detect_changes` re-derives this correctly, but it's work we don't need to own.
- **Much less code to maintain.** ~340 lines of custom watcher → ~50 lines using `watchdog.observers.Observer` + a debounced handler. Smaller surface area for bugs.

FeodorFitsner (approved):

PR #6403, opened 2026-04-10
fix(`flet pack`): Windows taskbar shortcut pointing to wrong `.exe`

### LINKED ISSUE #5151: Blank White Screen When Launching Flet App from Taskbar After Initial Launch
### Duplicate Check

### Describe the bug

After successfully launching a Flet App compiled with Flet Pack/PyInstaller/Nuitka for the first time, launching the app again via the taskbar shortcut created from the launched app results in a blank white screen. The shortcut seems to be incorrectly linking to C:\Users\(USERNAME)\.flet\bin\flet-0.25.2\flet\flet.exe, and running this results in a blank screen.

When dragging the compiled .exe to the taskbar and using it as a shortcut, the app opens with a new Flet window with the Flet icon, but it still targets the flet.exe located in C:\Users\(USERNAME)\.flet\bin\flet-0.25.2\flet\flet.exe, leading to the same issue.

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft

def main(page: ft.Page):
    page.add(ft.Text(value="Hello, world!"))

ft.app(main)
```

</details>

### To reproduce

1. Compile a Flet app using Flet Pack/PyInstaller/Nuitka.
2. Launch the compiled .exe successfully for the first time.
3. Create a shortcut on the taskbar by dragging the .exe to the taskbar or simply right clicking the launched app and hitting pin to taskbar.
4. Close the app.
5. Click on the taskbar shortcut to relaunch the app.

### Expected behavior

**Expected Behavior:**

The app should launch properly without showing a blank white screen when launched via the taskbar shortcut.

**Actual Behavior:**

Launching the app via the taskbar shortcut causes a blank white screen. The shortcut points to C:\Users\(USERNAME)\.flet\bin\flet-0.25.2\flet\flet.exe, which results in a failed launch.

### Screenshots / Videos

<details open>
<summary>Captures</summary>
The blank white screen after launching the shortcut:

Image

What the shortcut links to (which should link to the compiled .exe the compliler made):

Image

</details>

### Operating System

Windows

### Operating system details

Windows 11

### Flet version

flet 0.23.0 and higher have been tested and producing the same result

### Regression

I'm not sure / I don't know

### Suggestions

_No response_

### Logs

There is no useful logs as its not launching my application from the shortcut. If there is a location for logs that might be useful for my case, please tell me where I can find these and I will provide them.

</details>

### Additional details

This issue occurs only when launching the app from the taskbar after its first successful launch.

The taskbar shortcut appears to be pointing to the incorrect executable (flet.exe located in C:\Users\(USERNAME)\.flet\bin\flet-0.25.2\flet\flet.exe).

When the .exe that was created by flet pack/pyinstaller/nuitka is run directly, the issue does not occur.

ndonkoHenri:
- **Taskbar fix:** When a packed Flet app runs on Windows, it spawns `flet.exe` (the Flutter client) as a child process. Windows associates the taskbar entry with the child's executable path (`~/.flet/client/.../flet.exe`), so pinning the app creates a shortcut to the bare client. The fix sets a Windows `AppUserModelID` on `flet.exe` via the `FLET_APP_USER_MODEL_ID` environment variable, so Windows associates the taskbar entry with the parent application instead. Resolves #5151.
  Screenshot 2026-04-10 at 2 13 23 PM

- **Pack bundling fix:** `flet pack` was copying the *extracted* client files into the PyInstaller bundle, but at runtime `ensure_client_cached()` looks for an *archive* (`flet-windows.zip` / `flet-linux-*.tar.gz`). The archive was never found, so the app always fell back to downloading from GitHub on first launch — wasteful for released versions, broken for unreleased ones. macOS already re-packed correctly; Windows and Linux now do the same.

Testable CI artifacts: https://github.com/flet-dev/flet/actions/runs/24401327648

## Summary by Sourcery

Ensure packed Flet desktop apps bundle and launch the correct client archives across platforms and integrate with Windows taskbar pinning behavior.

Bug Fixes:
- Fix Windows taskbar pinning so packed Flet apps are associated with the parent executable instead of the cached flet.exe client.
- Ensure flet pack bundles the desktop client as an archive (zip/tar.gz) that matches ensure_client_cached() expectations on Windows and Linux, avoiding fallback network downloads.

Enhancements:
- Expose a public get_artifact_filename() helper for determining the platform-specific desktop client archive name.
- Improve pack command behavior by compressing patched client binaries into platform-appropriate archives before PyInstaller runs.
- Clarify desktop client and environment-variable documentation strings in flet_desktop.


FeodorFitsner:
Add changelog record.

sdk/python/packages/flet-desktop/src/flet_desktop/__init__.py
```diff
@@ -459,4 +458,11 @@ def __locate_and_unpack_flet_view(page_url, assets_dir, hidden):
     if hidden:
         flet_env["FLET_HIDE_WINDOW_ON_START"] = "true"
 
+    # On Windows, set AppUserModelID so the taskbar associates the Flet client window
+    # with the parent executable (e.g. a PyInstaller bundle) rather than the cached
```
FeodorFitsner: What if the client being run without PyInstaller?

ndonkoHenri: Good point. I have now moved it out of `flet_desktop`, into the PyInstaller runtime hook instead.

FeodorFitsner (approved):

PR #6463, opened 2026-05-03
feat(`flet-video`): support direct mutation of `playlist`, frame screenshots and video-controls configuration

ndonkoHenri:
- Add configurable `Video.controls` support for adaptive, Material, Material desktop, custom Flet controls, and hidden controls.
- Add `Video.take_screenshot()` for capturing video frames.
- Allow direct `Video.playlist` mutation for add/remove flows and deprecate `playlist_add()` / `playlist_remove()`.
- Add `Video.on_position_change`/`Video.on_duration_change` events.
- Add `parseControlWidget()` and `parseControlWidgets()` Dart-side helpers for `@value` dataclass objects that accept Flet controls.
- Fix control diffing for controls nested inside `@value` dataclass objects so they keep the nearest control parent/page context, and restore optional structured properties that are cleared to `None` and later set again

## Test Code

See examples.

## Summary by Sourcery

Add configurable video controls support, frame screenshot capture, and improved playlist mutation and diff handling for flet-video, along with tests and documentation updates.

New Features:
- Introduce configurable built-in and custom video controls through the Video.controls API, including adaptive, Material, and Material desktop control sets and button bar items.
- Add Video.take_screenshot() to capture the current video frame as PNG, JPEG, or raw BGRA bytes from Python.
- Enable direct mutation of Video.playlist for add/remove flows while deprecating legacy playlist_add() and playlist_remove() helpers.
- Expose Dart-side parseControlWidget() and parseControlWidgets() utilities for converting serialized Flet controls into Flutter widgets, and add rich video controls configuration parsing.

Bug Fixes:
- Fix diffing of controls nested inside @value dataclass objects so they retain the correct parent/page context and restore optional structural fields after being set to None and then re-populated.
- Ensure explicit empty lists and dicts on value objects are serialized through the protocol instead of being dropped as structural defaults.

Enhancements:
- Improve object diff builder typing, internal documentation, and handling of added/removed dataclasses and list reconciliation.
- Optimize video playlist updates by detecting single-item append/remove operations and applying targeted player updates instead of reopening the full playlist.

Documentation:
- Document new video controls types and button bar items and wire them into the sidebar navigation.
- Refresh video control examples to cover basic playback, playlists, events, subtitles, screenshots, controls selection, and button bar customization.
- Update project changelogs to describe the new flet-video capabilities and diffing fixes.

Tests:
- Extend dataclass and object-diff tests to cover nested controls in value objects, structural empty lists/dicts, and restoration of optional value fields after None.


bl1nch:
Hi @ndonkoHenri, thank you for looking into the issue and making a large PR.

Could I clarify one point? As far as I understand, in this implementation, changing the `playlist` property on the `Python` side will trigger the `didUpdateWidget` function on the `Flutter` side? If that is the case, then `_player.dispose()` will be called, followed by the creation of a new `_player = Player(configuration: playerConfig);` object.

As I understand it, this is a resource-intensive operation in media_kit which creates a new MPV instance and allocates resources (https://github.com/media-kit/media-kit/blob/7102e7da96f39c718487a8f7a59b6a034aae7f45/media_kit/lib/src/player/native/player/real.dart#L78, which is a heavy operation). The idea behind my `playlist_update` method was precisely to call `_player.open` on an already created player so that the video switch happens instantly, even while another video is playing; otherwise, one could recreate the `Video` control on the `Python` side in `Flet`, but that takes longer.

ndonkoHenri:
Hi @bl1nch. Thanks for raising this. Direct `playlist` mutation as implemented in this PR does not recreate the `Player`.

On the Dart side, `didUpdateWidget()` only calls `_teardown()` / `_setup()` when the `Control` instance changes. A Python-side `playlist` property patch updates the existing Dart `Control` and rebuilds it, so the existing `_player` remains alive.

The playlist update path is handled in `build()`:

- single append -> `_player.add(media)`
- single removal -> `_player.remove(index)`
- more complex replacement -> `_player.open(Playlist(...), play: autoplay || _player.state.playing)`

So this PR should be at least as lightweight as the `playlist_update()` approach for full replacement, and lighter for append/remove because it uses media_kit’s dedicated playlist mutation APIs.

bl1nch:
@FeodorFitsner, the changes in https://github.com/flet-dev/flet/pull/6463/changes/58d396fe32f36af566d4d4caecd64c11b0b50500 broke support for streaming protocols like `rtp://`, `rtsp://`, `rtmp://`, `udp://`, and `srt://`

Since `getAssetSource` only checks for `http`, `https`, `data`, and `blob`, it incorrectly treats other protocols as relative paths and prepends assets_dir to them

FeodorFitsner:
@bl1nch Fixed: https://github.com/flet-dev/flet/commit/47d5b5b01f70fa2a1d930dcfe4422bc1edab58e9

PR #6487, opened 2026-05-09
Prepare 0.85.1 release: flet-geolocator reliability fixes

FeodorFitsner:
## Summary

- Make `flet-geolocator.Geolocator` actually work end-to-end on web and stop silently swallowing errors on desktop:
  - `get_last_known_position()` now returns `Optional[GeolocatorPosition]` instead of crashing with `TypeError: argument after ** must be a mapping, not NoneType` when no fix is cached.
  - `get_current_position()` no longer hangs forever on web — added a Dart-side `Future.timeout` workaround for the upstream [`geolocator_web` 4.1.3](https://pub.dev/packages/geolocator_web) `inMicroseconds`/`inMilliseconds` timeout typo, and picked sensible web defaults (`time_limit: 30s`, `maximum_age: 5m`) that scope only to single-shot calls (the position stream is left alone).
  - Fixed the Dart side reading `args["settings"]` when Python sends `args["configuration"]`, which silently dropped the configuration on every `getCurrentPosition` call.
  - The position stream is now gated behind a registered `on_position_change` / `on_error` handler, and cancels the previous subscription on `update()` so configuration changes don't leak listeners or trigger CoreLocation prompts at random times.
  - Platform exceptions (`LocationServiceDisabledException`, `PermissionDeniedException`, `PermissionDefinitionsNotFoundException`, `PermissionRequestInProgressException`, `PositionUpdateException`, `TimeoutException`) are translated into actionable error messages via a private `_GeolocatorException` whose `toString()` drops the default `Exception:` prefix, so Python sees `RuntimeError("Location request timed out.")` rather than `RuntimeError("Exception: ...")`.
- Updated the example to wrap each `geo.*` call in `try/except RuntimeError` so failures land in a `SnackBar` instead of escaping to the global app-error banner, and to handle `None` from `get_last_known_position`.
- Added a docs section for the web cache window (`maximum_age` default + override) and a macOS troubleshooting note for the recurring `POSITION_UNAVAILABLE` failure mode (`sudo killall locationd`).
- Bumped `packages/flet/pubspec.yaml` to `0.85.1` and added CHANGELOG entries (root and `packages/flet/`) following the 0.83.1 precedent.

## Test plan

- [ ] Native macOS: Request Permission → Get Permission Status → Get Current Position returns a fix and renders a snackbar.
- [ ] Native macOS: Get Last Known Position returns either a position or "No last known position available." (no crash).
- [ ] Web (Chrome / Edge / Safari) on `localhost`: Get Current Position returns a fix when the browser can reach `locationd`; surfaces a snackbar with `Location request timed out.` (not a hang) when the OS pipeline is wedged.
- [ ] Web: position stream stays subscribed only when `on_position_change` is set; toggling the configuration cancels the previous subscription instead of leaking it.
- [ ] Permission-denied path: snackbar shows `Could not get current position: Location permission denied: ...` instead of a global error banner.
- [ ] Rebuild the Flutter web client (`client/build/web/`) and the desktop runner before regression-testing — Dart-side changes don't take effect otherwise.

## Summary by Sourcery

Improve flet-geolocator reliability and error handling across web and desktop and prepare the 0.85.1 release.

Bug Fixes:
- Prevent get_last_known_position from crashing when no cached position exists by returning an optional position value instead.
- Ensure get_current_position on web times out with a clear error instead of hanging indefinitely and respect web-specific configuration settings.
- Route configuration correctly from Python to the Dart geolocator calls and gate the position stream behind registered handlers to avoid leaks and spurious prompts.
- Normalize platform geolocation exceptions into user-friendly error messages surfaced to Python as RuntimeError.

Enhancements:
- Update the geolocator example app to wrap geolocation calls in error handling and surface issues through snackbars and logs.
- Introduce sensible default web cache and timeout behavior for single-shot geolocation requests and document these behaviors and macOS troubleshooting steps.
- Guard desktop settings-opening calls on web by returning a boolean result instead of doing nothing.

Build:
- Bump the flet Dart package version to 0.85.1 and add coordinated changelog entries for the core project and Dart package.

Documentation:
- Document web geolocation cache-vs-fresh behavior and configuration, and add macOS troubleshooting guidance for POSITION_UNAVAILABLE failures.


ndonkoHenri (approved):
lgtm

PR #6577, opened 2026-06-11
Multi-version Python support (3.12 / 3.13 / 3.14)

FeodorFitsner:
## Description

Adds multi-version bundled CPython support to `flet build` and `flet publish`. Users pick which interpreter ships with their app via a new `--python-version` flag (3.12 / 3.13 / 3.14) or via `[project].requires-python` in their `pyproject.toml`; the default is the latest supported stable (currently 3.14). Pyodide is no longer pre-baked into the build template — the matching release for each Python version is downloaded into the build output and cached under `~/.flet/pyodide/<version>/` per build.

Companion to [flet-dev/serious-python#207](https://github.com/flet-dev/serious-python/pull/207) (the 2.0.0 release where the platform plugins and packager learned to parameterize the bundled Python version). The build template here pins `serious_python: 2.0.0`, so this branch should land *after* serious_python 2.0.0 publishes to pub.dev.

### Highlights

- **New registry + resolver**: [`flet_cli/utils/python_versions.py`](sdk/python/packages/flet-cli/src/flet_cli/utils/python_versions.py) — single source of truth for Python ↔ CPython-standalone ↔ Pyodide ↔ wheel-platform-tag. `prerelease: True` rows are opt-in (skipped in open-ended `requires-python` resolution; reachable via explicit `--python-version 3.15` / `==3.15.*`).
- **`build_base.py` plumbing**: resolves once in `initialize_command`, threads `--python-version` + `SERIOUS_PYTHON_VERSION` into both the serious_python package step and the later `flutter build`, and exposes the resolved short version to the cookiecutter template as `cookiecutter.options.python_version` so the Android `abiFilters` can drop `armeabi-v7a` for 3.13+ (PEP 738).
- **Pyodide download + caching**: [`flet_cli/utils/pyodide.py`](sdk/python/packages/flet-cli/src/flet_cli/utils/pyodide.py) fetches `pyodide-core-<version>.tar.bz2` + `micropip` / `packaging` wheels into a per-version cache, idempotent.
- **`patch_index.py`**: injects `flet.pyodideUrl` per build (CDN or local `pyodide/pyodide.js`); the hardcoded `defaultPyodideUrl` constant is removed from both `client/web/python.js` and the build template `python.js`. Pre-baked pyodide bundle deleted from the build template.
- **`flet --version`**: lists supported Python versions (newest first) instead of a single static Pyodide line — the global `flet.version.pyodide_version` export is removed.
- **Docs**: \"Choosing a Python version\" section on the publish guide, Pyodide / Python pairing table on the static-website page, `--pre` vs `--python-version` clarification on `flet publish`.
- **CI**: `flet-build-test` now runs against the full 3.12 / 3.13 / 3.14 grid (14 build targets × 3 versions = 42 jobs), grouped via a reusable workflow so the Actions UI shows three cards instead of a flat 42-row list. `workflow_dispatch` gains a `python_version` choice input (`all` / `3.12` / `3.13` / `3.14`) for targeted manual runs.
- **Bugfix**: dev-package `file://` URLs on Windows now use \`Path.as_uri()\` so paths render as `file:///D:/...` instead of being misparsed as UNC.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue) — Windows dev-install `file://` URL
- [ ] Breaking change
- [x] This change requires a documentation update — included

## Checklist

- [x] I signed the CLA.
- [x] New and existing tests pass locally with my changes. (CI matrix in flet-dev/serious-python#207 covers the platform plugin side.)
- [x] I have made corresponding documentation changes, if applicable.
- [x] I have added changelog entries for user-facing changes, if applicable.
- [ ] I have updated release guide pages and `website/sidebars.yml` for breaking changes, removals, and deprecations, if applicable. (Not breaking from end-user perspective; existing builds keep working with `--python-version 3.12` if pinned.)

## Additional details

**Release sequencing**: this branch's build template pins `serious_python: 2.0.0`. Don't merge until serious_python 2.0.0 is live on pub.dev (flet-dev/serious-python#207).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Add configurable bundled Python version support to Flet builds and publishes, wiring version resolution through the CLI, web runtime, templates, and CI, and switch Pyodide to a per-version, cached download instead of a baked-in bundle.

New Features:
- Allow `flet build` and `flet publish` to select the bundled Python interpreter via a new `--python-version` flag or `project.requires-python`, with 3.12, 3.13, and 3.14 supported.
- Introduce a centralized Python version registry that maps supported CPython releases to their corresponding Pyodide versions and wheel platform tags.
- Emit a detailed supported-Python-versions listing from `flet --version`, including which version is the default and associated Pyodide releases.

Bug Fixes:
- Fix dev-package `file://` URLs on Windows by using `Path.as_uri()` so pip resolves drive-letter paths correctly instead of treating them as UNC paths.

Enhancements:
- Parameterize serious_python and Flutter build steps with the resolved Python version, exposing it to templates so Android ABIs match the selected runtime and PEP 738 constraints.
- Refine the web runtime so each build injects a Python-version-specific Pyodide CDN or local URL, removing hardcoded Pyodide defaults from the client and templates.
- Download and cache per-version Pyodide runtimes (including required wheels) under a shared cache directory, copying them into each web build or publish output as needed.

Build:
- Refactor the `flet-build-test` workflow into a reusable matrix workflow invoked once per supported Python version, and add a `workflow_dispatch` input to target specific Python versions while preserving full matrix coverage by default.

CI:
- Expand the build CI matrix to cover all supported bundled Python versions across platforms, grouping jobs by Python version for clearer visibility in GitHub Actions.

Documentation:
- Document how to choose the bundled Python version, including the CPython/Pyodide compatibility matrix, and clarify the distinction between `--pre` and `--python-version` in the publish docs.
- Update web publishing docs to describe how Python version selection affects Pyodide, and adjust examples and comparison tables to reference the new default Python requirement.


bl1nch:
@FeodorFitsner, excuse me, could you please clarify one point regarding the supported ABIs?

How difficult would it be on your end to keep supporting `armeabi-v7a` for all Python versions, including `3.13` and `3.14`? I'm not deeply familiar with the underlying technicalities, so I wanted to ask if this restriction is strictly mandatory. If these Python versions can be compiled for `armeabi-v7a` without making additional major changes, could you potentially support them as well?

`armeabi-v7a` is still very common, especially in low-end or portable devices like Android TV boxes. Even though most of these devices are built on processors that support 64-bit ARM instructions, manufacturers still use 32-bit operating systems to save memory. While I understand you are following the PEP, dropping `armeabi-v7a` support causes serious issues for these use cases, as there is currently no active trend toward moving to 64-bit OS for the devices mentioned above.

FeodorFitsner:
@bl1nch It's not difficult. We already control most of Python build process in https://github.com/flet-dev/python-build it's just [>3.12 branch](https://github.com/flet-dev/python-build/blob/main/android/build.sh#L149-L202) uses upstream CPython Android packaging suite which supports 64-bit arches only. We could just patch it and "enable" `armeabi-v7a`.

Let's look into it once the next Flet 0.86 is released? It's scope is huge.

For now you can just build against Python 3.12 which supports `armeabi-v7a` and we have commitment to support wheels for that arch as well.

PR #6591, opened 2026-06-15
fix(flet-secure-storage): bump flutter_secure_storage 10.0.0 -> 10.3.1 to fix macOS data-protection keychain (-34018) on unsigned/dev apps

### LINKED ISSUE #6586: flutter_secure_storage pinned to 10.0.0 — secure storage unusable on unsigned/dev macOS (darwin 0.2.0 bugs fixed upstream)
## Summary

`flet-secure-storage` pins `flutter_secure_storage: 10.0.0` in its bundled Flutter `pubspec.yaml`. That resolves `flutter_secure_storage_darwin` **0.2.0**, which has two macOS bugs — both already fixed upstream — that make secure storage unusable on **unsigned / development** macOS apps.

## The problem on macOS

On macOS, `flutter_secure_storage` defaults to the **data-protection keychain**, which requires the `keychain-access-groups` entitlement (an embedded provisioning profile). Unsigned or unprovisioned apps — including `flet run` and an unsigned `flet build macos` — fail every write with `errSecMissingEntitlement` (`-34018`).

The documented escape hatch is `MacOsOptions(usesDataProtectionKeychain: false)`, which should route to the **login keychain** (no entitlement required). In darwin 0.2.0 that option is **inert**:

1. **Key-name mismatch** — the Dart side sends `usesDataProtectionKeychain`, but the Swift side read a differently-cased key and defaulted it to `true`, so the data-protection keychain was always used regardless of the option.
2. A follow-up issue where `kSecUseDataProtectionKeychain` was added to the keychain query even when `false`, still tripping the entitlement.

Net effect with the current pin: there is **no way** for an unsigned/dev macOS Flet app to persist values — the data-protection path `-34018`s, and the login-keychain fallback is unreachable.

## Already fixed upstream

- Key-name fix — `flutter_secure_storage_darwin` **0.3.0** (juliansteenbakker/flutter_secure_storage#1047, merged 2026-02-10)
- Query fix — **0.3.2** (juliansteenbakker/flutter_secure_storage#1136, merged 2026-05-22)
- Current darwin is **0.4.0**; current umbrella `flutter_secure_storage` is **10.3.1** stable (`11.0.0-beta.1` prerelease requires `flutter_secure_storage_darwin: ^0.4.0`).

## Verification

I confirmed the fix end-to-end. A `flet build macos` app, **ad-hoc signed (no provisioning profile)**, with:

```toml
[tool.flet.flutter.pubspec.dependency_overrides]
flutter_secure_storage = "10.3.1"
flutter_secure_storage_darwin = "0.4.0"
```

and `MacOsOptions(uses_data_protection_keychain=False)` successfully wrote **and** read back a value from the macOS login keychain — no `-34018`. With the default pin (darwin 0.2.0) the same app fails. Inspecting the built framework, only the corrected `usesDataProtectionKeychain` key is compiled in once the override is applied.

Note: `flet run` uses the prebuilt `flet-desktop` binary, which bakes the plugin in at release time, so it cannot benefit from `dependency_overrides` and stays broken until a new flet-desktop binary ships against the newer plugin. A pin bump fixes both `flet build` and `flet run`.

Also worth flagging: on macOS, `get_availability()` returns `True` even while every write fails with `-34018`, so it can't be used as a pre-flight check — a write round-trip is the only reliable signal.

## Request

Bump `flet-secure-storage`'s `flutter_secure_storage` pin from `10.0.0` to a current release (latest stable `10.3.1`) that resolves `flutter_secure_storage_darwin ≥ 0.3.2`. This restores OS-backed secure storage for unsigned/dev macOS apps (via the login keychain) without requiring provisioning, and — once a new prebuilt binary ships — fixes `flet run` too.

Happy to open a PR for the pin bump if that would help.

## Environment

- `flet` / `flet-secure-storage` 0.85.2
- macOS (Apple Silicon)
- Pinned `flutter_secure_storage: 10.0.0` → resolves `flutter_secure_storage_darwin: 0.2.0`

EH-MLS:
## What

Bumps the `flutter_secure_storage` pin in the `flet-secure-storage` extension from `10.0.0` to `10.3.1`:

- `sdk/python/packages/flet-secure-storage/src/flutter/flet_secure_storage/pubspec.yaml` — the source-of-truth pin (`flutter_secure_storage: 10.0.0` → `10.3.1`).

I intentionally left the lockfiles out of this PR: `client/pubspec.lock` and `client/ios/Podfile.lock` need regenerating via `flutter pub get` (and `pod install`) so the transitive `flutter_secure_storage_darwin` 0.2.0 → ≥ 0.3.2 (currently 0.4.0) bump and its sha256 hashes land correctly — hand-editing them would produce an inconsistent lockfile. Happy to push the regenerated lockfiles, or for a maintainer to regenerate them in CI. **Allow edits by maintainers** is enabled.

## Why

The current `10.0.0` pin resolves `flutter_secure_storage_darwin 0.2.0`, which has a Dart→Swift key-name mismatch that makes `MacOsOptions(usesDataProtectionKeychain: false)` inert — the data-protection keychain is always used. On an unsigned/dev macOS app (no provisioning profile, so no `keychain-access-groups` entitlement) every write then fails with `errSecMissingEntitlement` (`-34018`), and the login-keychain escape hatch is unreachable. That makes it effectively impossible to persist app data/secrets while developing locally on macOS.

Fixed upstream in `flutter_secure_storage_darwin`:
- key-name fix in 0.3.0 (juliansteenbakker/flutter_secure_storage#1047)
- follow-up `kSecUseDataProtectionKeychain` fix in 0.3.2 (juliansteenbakker/flutter_secure_storage#1136)

Both ship transitively via `flutter_secure_storage 10.3.1` (current latest stable), so bumping the umbrella pin to `10.3.1` restores a working `MacOsOptions(usesDataProtectionKeychain: false)` and removes the `-34018` failure.

## Why bump the pin instead of using `dependency_overrides`

The `[tool.flet.flutter.pubspec.dependency_overrides]` workaround suggested in #6586 only fixes `flet build`, because that path compiles the Flutter client from source and re-reads the pubspec. It does **not** help `flet run`: the prebuilt `flet-desktop` / `flet-desktop-full` client bundles this extension (its framework ships inside the prebuilt app), built against the default `10.0.0` pin, so it still contains the buggy darwin 0.2.0 keys.

Bumping the single source-of-truth pin is the only change that fixes **both** paths: `flet build` (compiled from source) and `flet run` (the prebuilt client, once a new `flet-desktop` release ships with this pin).

## Verification

With `flutter_secure_storage 10.3.1` and `MacOsOptions(usesDataProtectionKeychain: false)`, an unsigned (ad-hoc-signed, no provisioning profile) macOS app successfully **wrote and read the login keychain** with no `-34018`. With the default `10.0.0` pin the same app fails on every write.

## Notes

- Fixes #6586.
- Happy to add a `CHANGELOG.md` entry and/or retarget to the active `release/v{version}` branch if you'd prefer — just let me know.
- I can push the regenerated lockfiles, or hand this off for a maintainer to drive the lockfile regeneration in CI.

## Summary by Sourcery

Bug Fixes:
- Resolve macOS keychain write failures on unsigned/dev apps by updating flutter_secure_storage to a version that includes the fixed darwin implementation.

### DIFF
--- a/sdk/python/packages/flet-secure-storage/src/flutter/flet_secure_storage/pubspec.yaml
+++ b/sdk/python/packages/flet-secure-storage/src/flutter/flet_secure_storage/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   flutter:
     sdk: flutter
 
-  flutter_secure_storage: 10.0.0
+  flutter_secure_storage: 10.3.1
 
   flet:
     path: ../../../../../../../packages/flet


EH-MLS:
Retargeted this to `flet-0.86` so it can ride the current release cycle (since `main` hasn't moved). `flet-0.86` is only one commit ahead of `main`, and that commit doesn't touch this extension's `pubspec.yaml`, so the diff is unchanged (still the single pin line) and it merges cleanly. Happy to move it back to `main` if you'd prefer — edits-by-maintainers is on either way.

PR #6624, opened 2026-06-30
Add flet-mcp: MCP server for LLM agents

FeodorFitsner:
## Summary

Adds the **flet-mcp** package — an MCP server that gives LLM agents Flet API, icons, examples, docs, and CLI lookups — rebased onto `flet-0.86` and verified end to end.

Tool groups gate on `FLET_MCP_ENABLE_*` (API + icons on by default; examples, docs, CLI off).

## What works (verified)
- **API** — `api.json` built via Griffe against the 0.86 source: 249 controls, 91 events, 201 types, 149 enums, 13 CLI commands. Spot-checked.
- **Icons** — `find_icon` with synonym matching.
- **Examples** — 519 indexed; search ranking, `get_example` round-trip, and platform filter all validated against the current `[tool.flet.metadata]` / `[tool.flet].platforms` layout.
- **CLI** — `get_cli_help` from the `flet_cli` parser.
- **In-process use** — `from flet_mcp import mcp` + `fastmcp.Client(mcp)` exercised across all 11 tools.

## Docs tools — deferred
The docs indexer parses a mkdocs `search_index.json` that the site no longer produces after the Docusaurus + Algolia migration. The `DOCS` group stays **off** and now **degrades gracefully** (table guards return empty results instead of crashing). README and the `--docs` CLI help note the deferral. Rebuilding docs search against Docusaurus is tracked as follow-up.

## CI / packaging
- New `build_flet_mcp` job builds the indexes (`flet mcp build --examples`) then `uv build --package flet-mcp`.
- Feeds the `release` job and `flet_mcp` is included in the `py_publish` loop.
- Verified the built wheel bundles `data/api.json` and `data/mcp.db` (via `package-data = data/*`).

## Notes
- `api.json` and `mcp.db` are gitignored build artifacts, generated at packaging time.

## Summary by Sourcery

Introduce a new flet-mcp MCP server package for Flet, integrate its build and publication into the Python SDK and CI, and add metadata hooks for richer API indexing and MCP discovery.

New Features:
- Add the flet-mcp package providing an MCP server with tools for Flet API, icons, examples, docs, and CLI integration.
- Expose an MCP-focused CLI command `flet mcp` with build and serve subcommands and optional in-process usage via FastMCP.
- Add optional MCP metadata (categories, tags) to controls like TextField, Column, Row, Checkbox, and Tabs for better discovery.

Enhancements:
- Extend gradient color handling to accept generic color values instead of plain strings.
- Clarify and slightly expand color usage documentation, including named colors and opacity behavior.
- Wire flet-mcp as a workspace member and dependency in the Python SDK, including a dedicated mcp-build dependency group and isort config updates.

Build:
- Add a dedicated build_flet_mcp CI job that builds MCP data, packages the flet-mcp wheel, and uploads artifacts.
- Include flet-mcp in the release and publishing workflows so it is built and published alongside other Python packages.

Documentation:
- Tweak colors cookbook docs for formatting and to clarify named color identifiers and opacity helper usage.


sdk/python/packages/flet/src/flet/controls/base_control.py
```diff
@@ -103,25 +107,41 @@ def control(
         isolated: If `True`, marks the control as isolated. An isolated control
             is excluded from page updates when its parent control is updated.
         post_init_args: Number of InitVar arguments to pass to __post_init__.
+        categories: MCP metadata — control categories (e.g. ``("input", "form")``).
```
ndonkoHenri: single backsticks instead of double

sdk/python/packages/flet/src/flet/controls/material/tabs.py
```diff
@@ -162,7 +162,7 @@ class UnderlineTabIndicator:
     """
 
 
-@control("Tabs")
+@control("Tabs", categories=("navigation", "layout"), tags=("tabbed", "switcher"))
```
ndonkoHenri, line 165: Hum, so we will have to do this for all controls?

sdk/python/packages/flet-mcp/README.md
ndonkoHenri, line 1: will be nice to show this in documentation too - just like we do for extensions atm.
from what i see we currently only display the  cli help.

sdk/python/packages/flet/src/flet/controls/material/tabs.py
FeodorFitsner: it's experimental. As AI is a non-deterministic, we currently don't know what would really help it to do better decisions. If these categories help - we will annotate all controls; if it happens to be a "noise" - we'll remove.

Issue #86, opened 2022-07-20
`DatePicker` control


ogustbiller:
Do you have plans aboou DatePicker control  realisation?

FeodorFitsner:
Like this one: https://api.flutter.dev/flutter/material/showDatePicker.html?

ogustbiller:
Yes, for example.

clstaudt:
Can't wait to replace homemade controls like this... 

```python
class DateSelector(UserControl):
    """Date selector."""

    def __init__(self):
        super().__init__()

        self.day_dropdown = Dropdown(
            label="D",
            options=[
                dropdown.Option(day) for day in range (1, 32)
            ],
            width=50,
        )

        self.month_dropdown = Dropdown(
            label="M",
            options=[
                dropdown.Option(month) for month in range (1, 13)
            ],
            width=50,
        )

        self.year_dropdown = Dropdown(
            label="Y",
            options=[
                dropdown.Option(year) for year in range (2015, 2025)
            ],
            width=100,
        )


        self.view = Container(
            content=Row(
                [
                    Icon(
                        icons.CALENDAR_MONTH,
                    ),
                    self.day_dropdown,
                    self.month_dropdown,
                    self.year_dropdown,
                ],
                alignment="center",
            ),
            padding=10,
        )

    def build(self):
        return self.view

    def get_date(self) -> datetime.date:
        """Return the selected timeframe."""
        date = datetime.date(
            year=int(self.year_dropdown.value),
            month=int(self.month_dropdown.value),
            day=int(self.day_dropdown.value),
        )
        return date
```

ogustbiller:
> Can't wait to replace homemade controls like this...
> 

Thank you! 🙏.  Very nice

clstaudt:
@ogustbiller   Not nice if you ask me, just a makeshift solution - which took too long to build - until the DatePicker control arrives.

clstaudt:
Can we have an update please? It's a control type that is needed frequently.

ckaldemeyer:
+1

clstaudt:
@vlad-et-git FYI

ipritom:
+1

joeg-ita:
https://github.com/joeg-ita/flet-datepicker

ndonkoHenri:
@FeodorFitsner will you finally import that made by @beret in one PR? Or probably bring it in yourself?

FeodorFitsner:
[...]

PR #6673, opened 2026-07-12
Raw RGBA Matplotlib frames over DataChannels; canvaskit web default; restore flet-charts tests

FeodorFitsner:
## Summary

Fast Matplotlib rendering path: when the client runs on the same machine (embedded native, Pyodide, dev socket), `MatplotlibChart` now streams uncompressed RGBA full frames straight from Agg's buffer over the DataChannel — no numpy diffing, no PNG encode/decode — displayed with a single `decodeImageFromPixels` + swap + dispose. Remote WebSocket clients keep the compact PNG full+diff pipeline (auto-selected per connection via a new `Connection.local_data_transport` capability flag).

### Wire format

New opcode on the existing widget DataChannel: `[0x04][width u32 LE][height u32 LE][premultiplied RGBA8888]`. Ack/backpressure unchanged; unknown opcodes are now acked so a newer Python never hangs an older client.

### Transport fixes found along the way

- **O(n²) packet reassembly** in the Dart socket transport: the inbound buffer was flattened on every incoming chunk; multi-MB frames arrive in dozens of chunks. Now flattens once per completed packet.
- Single-copy DataChannel frame framing in `FletSocketServer`.

### Benchmarks (three_d figure, 1600×1000 logical @ DPR 2 → 24 MB/frame, local socket)

| pipeline | mpl render+capture | transport+display | fps |
|---|---|---|---|
| PNG full+diff (before) | 106.8 ms | 28.8 ms | 7.4 |
| Raw frames (after) | 32.7 ms | 20.9 ms | **18.7** |

matplotlib's own render is now the dominant per-frame cost.

### canvaskit web renderer default

With `auto`, Chromium browsers select the dart2wasm/skwasm build, where every JS ↔ Dart byte-buffer crossing pays a WasmGC boundary conversion instead of a memcpy. Measured on a Pyodide app: 67 ms vs 10 ms per-frame display (Edge/skwasm vs Safari/canvaskit fallback) for the same 24 MB raw frames. Flet web apps exchange bytes with the Python runtime on every UI update, so `flet build web` / `flet publish` now default to `canvaskit` (override via `--web-renderer` or `[tool.flet.web].renderer`, which `flet publish` previously ignored due to an argparse default shadowing it). Documented in the publish guide, CLI help and `WebRenderer` docstrings.

### Restored flet-charts integration tests

The suite had been failing at collection since the examples reorg deleted the modules it imported — golden regressions went undetected. Tests now load the relocated example apps by path, all macOS goldens are regenerated and visually verified, and the matplotlib tests exercise the raw frame pipeline on every run. Reviving the suite immediately surfaced 12 broken chart examples (`SafeArea` wrappers missing `expand=True` → blank or shrunken charts), all fixed. `*_actual.png` diagnostic screenshots are now gitignored.

## Test plan

- `pytest packages/flet-charts/integration_tests/examples/` — 24/24 green (localhost socket → raw 0x04 path + goldens)
- Manual: `flet run` desktop (raw), `flet run --web` (PNG fallback), Pyodide build in Safari + Edge (raw; fps parity after canvaskit default)
- `flutter test` transport tests, `flutter analyze`, `ruff` clean

## Summary by Sourcery

Introduce a fast raw RGBA Matplotlib rendering path over local DataChannels, switch Flet web builds to use CanvasKit by default for better byte-buffer performance, and restore chart integration tests to cover the reorganized examples and golden screenshots.

New Features:
- Add support for streaming premultiplied raw RGBA matplotlib frames over DataChannels for local transports, alongside the existing PNG full/diff pipeline.
- Expose MatplotlibChartCanvas helpers on Python and Dart sides to encode, send, parse, and render raw frame packets.
- Provide example and stats tooling in the 3D matplotlib chart demo to measure raw vs PNG frame sizes and performance.

Enhancements:
- Optimize socket backend packet reassembly and DataChannel framing to avoid quadratic copying and reduce payload concatenations.
- Mark same-machine transports (desktop socket server, Dart bridge, Pyodide) with a local_data_transport flag so widgets can choose uncompressed payloads.
- Update chart examples to use SafeArea/Column expand flags to ensure charts render correctly and fill the available space.
- Adjust WebRenderer semantics and CLI defaults so web builds and publishes prefer CanvasKit, improving JS↔Dart byte buffer throughput on typical Flet apps.

Build:
- Change flet build and publish CLI defaults to use the CanvasKit renderer when no explicit web renderer is provided, and fix argument handling so pyproject settings are respected.

Documentation:
- Revise web static site publishing docs to document CanvasKit as the default renderer, its performance characteristics, and the precedence of web renderer settings.
- Add Android breaking-change guides explaining zipped site-packages with extract_packages and the removal of x86 from supported architectures, and link them from the breaking changes index.

Tests:
- Restore and update flet-charts integration tests to load relocated example apps by path and capture full-page screenshots where needed.
- Switch chart integration tests to the new example loader module and regenerate macOS golden images, ensuring matplotlib and other chart types are covered.
- Gitignore diagnostic *_actual.png screenshots produced by golden tests to avoid polluting the repo.

Chores:
- Add sidebar entries and cross-links for new Android breaking-change docs in the website navigation.


website/docs/updates/breaking-changes/v0-86-0/android-extract-packages.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Migration guide
+
+### List path-hungry packages in `extract_packages`
+
+Add the failing packages to your `pyproject.toml`:
+
+```toml
+[tool.flet.android]
+extract_packages = ["matplotlib", "scikit-learn"]
```
ndonkoHenri: From my testing, the entry is the package's **import name** — its top-level directory in site-packages — not the PyPI/distribution name. So `sklearn`, not `scikit-learn`; `cv2`, not `opencv-python`. (`matplotlib` matches only because its import name happens to equal its PyPI name)

```suggestion
extract_packages = ["matplotlib", "sklearn"]
```

Other packages that need an entry:

| package | entry | why |
|---|---|---|
| `astropy` | `"astropy"` | reads `astropy/CITATION` via `__file__` at import |
| `thinc` | `"thinc"` | reads `thinc/backends/_custom_kernels.cu` via `__file__` at import |
| `spacy` | `"spacy", "thinc"` | imports thinc at load (so it hits thinc's `_custom_kernels.cu`) and reads its own lang data via `__file__` — list both |
| `opencv-python` | `"cv2"` | cv2's bootstrap resolves its config files and loads its native extension through `__file__`-relative paths, so it must ship as a real directory |

website/docs/updates/breaking-changes/v0-86-0/android-x86-arch-removed.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+flet build apk --arch arm64-v8a x86_64
+```
+
+Nothing is lost by dropping it — previous builds never shipped a working x86 app. If you were
+targeting x86 for Android *emulators*, use an `x86_64` (or `arm64-v8a`) emulator system image
+instead; x86 images are only published for long-obsolete API levels.
+
+Note that the architectures available for a build also depend on the
+[bundled Python version](../../../publish/index.md#bundled-python) — see
```
ndonkoHenri: `#bundled-python` exists? Checked on my end, but cant find it. 

Also, the comment tries to point to a section (at least, from my understanding) which could be showing "android abi against python version table", which we dont currently have. Could be worth adding such a table or mentioning in some other form which abis each python version supports.

website/docs/updates/breaking-changes/v0-86-0/android-extract-packages.md
FeodorFitsner: Fixed in 310183249 — examples now use `sklearn`, added an explicit note that entries are import names (top-level directory under site-packages) rather than PyPI distribution names, and folded your tested packages (cv2, astropy, thinc, spacy+thinc) into the known-packages table. Thanks for testing these!

website/docs/updates/breaking-changes/v0-86-0/android-x86-arch-removed.md
FeodorFitsner: Good catch — the anchor never existed; the section is `#choosing-a-python-version` (same broken link was in publish/android.md too). Fixed both in 310183249 and inlined an ABI × Python version support matrix in the x86 guide, complementing the per-ABI list in [android.md#supported-target-architectures](https://flet.dev/docs/publish/android#supported-target-architectures). Also fixed a typo'd NDK anchor (#86-64 → #x86-64) while there.

FeodorFitsner: Correction in 508a48b83: the ABI matrix I added (and the pre-existing "armeabi-v7a — Python 3.12 only" / PEP 738 note in android.md it was based on) was wrong — flet-dev/python-build publishes all three ABIs including armeabi-v7a for every supported Python version (verified against the 20260708 manifest). Both docs now state the same thing: all three ABIs, every bundled Python version; upstream's PEP 738 32-bit drop doesn't apply to Flet's own CPython builds.

PR #6747, opened 2026-08-03
fix(testing): prevent host Python env from leaking into Flutter tests

PythBuster:
## Description

FletTestApp started the Flutter integration-test process with the complete environment inherited from the host Python process.

IDEs such as PyCharm add debugger and sitecustomize directories to PYTHONPATH. That value was inherited by `flutter test` and subsequently passed to the embedded Serious Python runtime.

As a result, the packaged test app could exit before connecting to RemoteTester. Flutter then reported exit code 79 with "No tests were found", while the flet_app fixture failed during setup.

Create an explicit environment for the Flutter subprocess and remove PYTHONPATH and PYTHONHOME before launching it.  All Flet, Flutter and SERIOUS_PYTHON_* variables remain unchanged.

This keeps host-only Python configuration out of the packaged runtime while preserving the environment required to build and execute integration tests.

Verified with the PyCharm TeamCity pytest plugin and PyCharm helper paths present in the host PYTHONPATH.

## Test code

```python
# Minimal test/reproduction code for reviewers, if applicable.
```

## Type of change

## Checklist

- [x] I signed the CLA.
- [ ] My code follows the style guidelines of this project.
- [x] I have made corresponding documentation changes, if applicable.

## Summary by Sourcery

Bug Fixes:
- Prevent host IDE and debugger Python configuration from causing packaged Flutter integration tests to exit prematurely by stripping PYTHONPATH and PYTHONHOME from the Flutter test environment.

### DIFF
--- a/sdk/python/packages/flet/src/flet/testing/flet_test_app.py
+++ b/sdk/python/packages/flet/src/flet/testing/flet_test_app.py
@@ -314,11 +314,19 @@ class FletTestApp:
                         f"--dart-define=FLET_TEST_ASSETS_DIR={self.__assets_dir}"
                     ]
 
+        # Do not leak host-Python configuration into Flutter's embedded Python
+        # IDEs such as PyCharm add debugger/sitecustomize modules through
+        # PYTHONPATH, which can break the packaged integration-test app.
+        flutter_env = os.environ.copy()
+        flutter_env.pop("PYTHONPATH", None)
+        flutter_env.pop("PYTHONHOME", None)
+
         self.__flutter_process = await asyncio.create_subprocess_exec(
             *flutter_args,
             cwd=str(self.__flutter_app_dir),
             stdout=stdout,
             stderr=stderr,
+            env=flutter_env,
         )
 
         if self.__flutter_process.stdout is not None:


FeodorFitsner:
Optional follow-up, same rationale as this fix - not blocking.

`PYTHONPATH`/`PYTHONHOME` are the two that bit you here, but they aren't the only host-Python knobs the embedded interpreter reads at init. Two more in the same family:

- **`PYTHONNOUSERSITE`** - worth *setting* to `1` rather than popping. If the host's `~/.local/lib/pythonX.Y/site-packages` happens to match the embedded interpreter's version, `site.py` adds it to `sys.path` and host packages leak in exactly the way this PR is preventing. Setting it explicitly is what actually blocks that, since the default (unset) means user site is enabled.
- **`PYTHONEXECUTABLE`** - macOS framework builds use it to seed `sys.executable`; a host value pointing at the IDE's interpreter is wrong for the packaged app.

Both are pre-existing hazards rather than anything this PR introduces, so they're fine as a separate change. If they do get added, it might be worth flipping the shape from a denylist of pops to a small helper that builds the env - something like:

```python
def _flutter_subprocess_env() -> dict[str, str]:
    """
    Environment for the `flutter test` child. Host-Python configuration must
    not reach the app's embedded interpreter: IDEs (PyCharm) inject debugger
    and sitecustomize paths via PYTHONPATH, which the packaged app imports at
    startup and dies on - surfacing as Flutter exit code 79, "No tests were
    found", before RemoteTester ever connects. All FLET_*, FLUTTER_* and
    SERIOUS_PYTHON_* variables are preserved.
    """
    env = os.environ.copy()
    for key in ("PYTHONPATH", "PYTHONHOME", "PYTHONEXECUTABLE"):
        env.pop(key, None)
    env["PYTHONNOUSERSITE"] = "1"
    return env
```

That also makes the one testable part of this - which keys are dropped, which survive - unit-testable without launching Flutter, and it parallels `_flutter_path_env` in `flet_cli/commands/test.py`.

FeodorFitsner (approved):

Issue #6755, opened 2026-08-06
bug: InteractiveViewer does not block scroll event from bubbling up to parent


eliferin14-svg:
### Duplicate Check

- [x] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

When I have an InteractiveViewer inside a scrollable element, like a Column, if I use the mouse wheel to zoom in the viewer it also scrolls the column

In the script pasted I tried to implement a callback (only on the first image) to disable column scrolling when I hover the viewer, but it just means that the column scroll state is reset to the zero position, and it is not an acceptable solution. On the second image you can see that you both zoom in/out and scroll the column

I am running flet on a Windows PC

### Code sample

<details open><summary>Code</summary>

```python
import flet as ft

URL = "https://picsum.photos/800/600"


@ft.component
def ImageComparison():
    is_hovered, set_is_hovered = ft.use_state(False)
    print(is_hovered)

    def handle_enter(e: ft.HoverEvent):
        set_is_hovered(True)

    def handle_exit(e: ft.HoverEvent):
        set_is_hovered(False)

    return ft.Column(
        expand=True,
        scroll=None if is_hovered else ft.ScrollMode.AUTO,
        spacing=30,
        controls=[
            ft.Row(
                controls=[
                    ft.Text(
                        "Interactive with scroll lock when hovering",
                        weight=ft.FontWeight.BOLD,
                        width=120,
                    ),
                    ft.Container(
                        clip_behavior=ft.ClipBehavior.HARD_EDGE,
                        content=ft.GestureDetector(
                            # on_hover=handle_hover,
                            on_enter=handle_enter,
                            on_exit=handle_exit,
                            content=ft.InteractiveViewer(
                                content=ft.Image(
                                    src=URL,
                                    fit=ft.BoxFit.CONTAIN,
                                ),
                            ),
                        ),
                    ),
                ],
                vertical_alignment=ft.CrossAxisAlignment.CENTER,
            ),
            ft.Row(
                controls=[
                    ft.Text("Interactive", weight=ft.FontWeight.BOLD, width=120),
                    ft.Container(
                        clip_behavior=ft.ClipBehavior.HARD_EDGE,
                        content=ft.InteractiveViewer(
                            content=ft.Image(
                                src=URL,
                                fit=ft.BoxFit.CONTAIN,
                            ),
                        ),
                    ),
                ],
[...]

PR #6761, opened 2026-08-07
fix: add listener widget for interactive viewer

7576457:
## Description

* gh-6755

## Test code

```python
import flet as ft


@ft.component
def ImageComparison():
    return ft.Container(
        bgcolor='blue',
        expand=True,
        content=ft.Column(
            scroll=ft.ScrollMode.AUTO,
            expand=True,
            alignment=ft.MainAxisAlignment.CENTER,
            controls=[
                ft.Container(),
                ft.Container(
                    bgcolor=ft.Colors.random(),
                    content=ft.InteractiveViewer(
                        content=ft.Image(
                            src="https://picsum.photos/800/400",
                        ),
                    )
                ),
                ft.Container(
                    bgcolor=ft.Colors.random(),
                    content=ft.InteractiveViewer(
                        content=ft.Image(
                            src="https://picsum.photos/800/500",
                        ),
                    )
                ),
                ft.Container(
                    bgcolor=ft.Colors.random(),
                    content=ft.InteractiveViewer(
                        content=ft.Image(
                            src="https://picsum.photos/800/600",
                        )
                    )
                )
            ]
        )
    )


def main(page: ft.Page):
    page.title = "Image Comparison"
    page.render(ImageComparison)


ft.run(main)

```

## Type of change

## Checklist

- [x] I signed the CLA.
- [ ] I have commented my code, particularly in hard-to-understand areas.

## Video

[Screencast From 2026-08-07 15-25-26.webm](https://github.com/user-attachments/assets/d1a4d097-4705-4968-8419-1b679ac6c9ef)

## Summary by Sourcery

Prevent InteractiveViewer pointer interactions from unintentionally scrolling parent containers while preserving parent scrolling for signals the viewer does not consume.

Bug Fixes:
- Prevent pointer signals handled by InteractiveViewer from also scrolling an enclosing scrollable.
- Allow pointer signals through when the viewer is disabled for that interaction or when its transform cannot change at the current boundary.


FeodorFitsner:
Reviewed - the approach is correct and I verified the mechanism works. A few things before merge.

### Why the bug happens (for the record)

Flutter's `InteractiveViewer` handles pointer signals directly in its own `Listener` callback and never registers with `GestureBinding.instance.pointerSignalResolver` (`_receivedPointerSignal`, `widgets/interactive_viewer.dart` in 3.44.8 - there is no `pointerSignalResolver` reference anywhere in that file). `Scrollable` *does* register (`widgets/scrollable.dart:962`). Since nothing claimed the event, the ancestor `Scrollable` wins by default, so the wheel zooms *and* scrolls. This PR claims the event with a no-op callback so `Scrollable`'s later `register()` is discarded (first registration wins).

I confirmed both halves with a throwaway widget test - a `ListView` at offset 400 with the viewer in it, one mouse wheel tick over the viewer:

| | scroll offset | scale |
|---|---|---|
| without the wrapper | 400 -> **350** (the bug) | 1.284 |
| with the wrapper | **400** (fixed) | 1.284 |

Zoom keeps working precisely because `InteractiveViewer` never went through the resolver to begin with.

I also diffed `_handlesPointerSignal` against Flutter 3.44.8's `_receivedPointerSignal` branch by branch, and it mirrors it correctly: trackpad without `trackpadScrollCausesScale` -> pan branch gated on `pan_enabled`, `scrollDelta.dy == 0` ignored, otherwise scale gated on `scale_enabled`, `PointerScaleEvent` -> scale. Defaults match the ones passed in `build`. Scrollables nested *inside* the viewer are unaffected, since they hit-test deeper and register first.

### Requested changes

1. **Add a comment explaining the trick.** A callback that deliberately does nothing is the least self-evident code there is - without a sentence saying "Flutter's InteractiveViewer handles pointer signals without claiming them via the PointerSignalResolver, so an ancestor Scrollable scrolls too; register a no-op to claim the event first", the next reader will delete it as dead code. Every other method in this file carries a `///` doc.

2. **CHANGELOG entry** - this is user-facing behavior and the checklist item is unchecked.

3. Minor: the three property reads are duplicated from `build`, so their defaults now have to be kept in sync by hand. Hoisting them into fields set in `build` removes the drift risk.

### One design question

The predicate claims the event whenever the viewer *would* handle it, not whenever it actually *did* something. That creates two dead zones:

* **Trackpad on macOS**: with the default `constrained=True` and a fitted image at scale 1.0, a two-finger scroll pans nothing (the matrix clamps at the boundary), but the event is now claimed, so the page does not scroll either. Before this PR the page at least scrolled. That is a regression for trackpad users on exactly the layout from the issue - an image inside a scrolling column.
* **Wheel at max/min scale**: same shape. Zoom cannot go further and the page will not scroll.

There is a clean fix, and I verified the ordering it relies on: the wrapper `Listener` fires *after* `InteractiveViewer` has already applied its transform (in a `maxScale: 4.0` viewer the wrapper observed scale 1.284, i.e. post-transform). So the handler can keep the previous `_transformationController.value` and only register when the matrix actually changed - the viewer consumes the event when it does something and hands it to the parent scrollable when it does not.

Whether you want that is a judgment call, since some viewers deliberately swallow at the zoom limit, but the trackpad case looks worth handling either way.

### CI

`examples/apps Integration Tests` failed at exactly `1h0m11s` with no `*_actual.png` artifacts, which is a job timeout rather than a screenshot diff, and `zizmor` is a workflow linter - neither can be affected by this diff. The branch is from Aug 7 though, so please rebase on `main` and re-run to get the checks green.

FeodorFitsner (approved):

PR #6773, opened 2026-08-16
refactor!: Replace `InputBorder` enum with a class hierarchy

ndonkoHenri:
`InputBorder` was an enum (`OUTLINE`/`UNDERLINE`/`NONE`) paired with five loose properties on every form field: `border_radius`, `border_width`, `border_color`, `focused_border_width` and `focused_border_color`. That shape could not express Flutter's API — `gap_padding` was unavailable, `UnderlineInputBorder`'s corner radius was silently ignored, and the error and disabled borders could not be styled at all — and every new Flutter border property would have required another top-level property on each control.

`InputBorder` is now a base class with `OutlineInputBorder`, `UnderlineInputBorder` and `InputBorder.none()`, mirroring Flutter's classes and their defaults. `border` accepts either a single border or a `ControlState` dictionary, so the focused, error and disabled borders become stylable.

```python
# before
ft.TextField(border=ft.InputBorder.UNDERLINE)
ft.TextField(
    border_radius=30,
    border_color=ft.Colors.GREEN_800,
    focused_border_color=ft.Colors.GREEN_ACCENT_400,
    focused_border_width=5,
)

# after
ft.TextField(border=ft.UnderlineInputBorder())
ft.TextField(
    border={
        ft.ControlState.DEFAULT: ft.OutlineInputBorder(
            border_radius=30,
            side=ft.BorderSide(color=ft.Colors.GREEN_800),
        ),
        ft.ControlState.FOCUSED: ft.OutlineInputBorder(
            border_radius=30,
            side=ft.BorderSide(width=5, color=ft.Colors.GREEN_ACCENT_400),
        ),
    },
)
```

Migration guide: `website/docs/updates/breaking-changes/v1-0-0/inputborder-class-hierarchy.md`.

## Nothing is removed in 1.0.0

The old spellings keep working for three minor releases, so existing apps run unedited:

- `InputBorder.OUTLINE`, `UNDERLINE` and `NONE` still resolve. A metaclass serves them, returning the equivalent class instance — so old code not only runs but produces new-API values, and `InputBorder.OUTLINE == OutlineInputBorder()` holds.
- `border_radius`, `border_width`, `border_color`, `focused_border_width` and `focused_border_color` still apply on `TextField`, `Dropdown`, `DropdownM2` and `CupertinoTextField`.
- `DropdownM2.border_radius` is an alias for the new `menu_border_radius`.

All of them warn through `V.deprecated` and are scheduled for removal in `1.3.0`.

Old and new combine through the repo's non-copying model: Python sends both, Dart prefers the new one. A `border` carrying a side or per-state entries ignores the deprecated properties entirely, while a bare `border=ft.InputBorder.UNDERLINE` still picks up a legacy `border_color`. This is why `border` is `Optional` again — with a `default_factory` the client cannot distinguish a set value from a default, so the fallback could never fire. It becomes non-optional when the deprecations are removed.

## What is still breaking

Only what a shim cannot cover:

- **Enum-shaped usage.** `InputBorder` is not iterable, its members have no `.value` or `.name`, and `InputBorder.OUTLINE is InputBorder.OUTLINE` is now `False` — each access returns a new instance, so compare with `==`.
- **Rendering.** A border with no explicit `side` takes its colour and weight from the Material theme per state instead of always painting black, so dark mode and custom themes work. An underline finally honours its `border_radius`. `DropdownM2`'s open menu is shaped by `menu_border_radius` rather than the field's radius. On `CupertinoTextField`, `InputBorder.none()` now actually removes the border (the enum value was ignored) and an outline without a `side` keeps the native iOS one.
- **Property reads.** Properties whose Flutter default is a fixed constant now declare it rather than `Optional[...] = None`, so reading one returns the value the control applies instead of `None`: the eight `Paint` style properties, `RoundedRectangleBorder.radius`, `Button.autofocus`, `Text.no_wrap`, `GridView.clip_behavior`, `Semantics.container`, `ExpansionPanelList.spacing`, `TextField.fit_parent_size`, `Page.show_semantics_debugger`, the three `CupertinoAppBar.automatic*` flags, `Path.Rect.border_radius` and `canvas.Text.max_width`. Rendering is unchanged, and properties a widget resolves at runtime from the theme, the platform or its own state keep `None`.

## Also in this PR

**A protocol constraint recorded in `protocol.py`.** The encoder emits nested dataclasses unconditionally, including when they equal their field's `default_factory` product, while the list, dict and scalar branches beside it prune. That asymmetry is load-bearing: the differ patches nested fields in place with nested-path ops, which require the client to already hold the parent key, so pruning is only safe alongside a differ that emits whole-value replaces for pruned fields.

## Docs and examples

- New breaking-change guide, plus the `InputBorder` type page split into a union index with `OutlineInputBorder` and `UnderlineInputBorder` pages. `NoInputBorder` is intentionally absent — the class is private and reached through `InputBorder.none()`.
- New examples: `types/input_border/showcase` (the three border styles), `types/input_border/styling` (custom sides, radii and per-state borders), and `material/dropdownm2/styling` (field border vs menu radius — `DropdownM2` had no examples before).
- Release notes gain a 1.0.x section and the missing 0.86.x entries. The changelog separates the breaking changes from the deprecations.

## Summary by Sourcery

Adopt a Flutter-aligned input border class hierarchy with state-aware styling while providing a migration path for the deprecated enum and border properties.

New Features:
- Replace the enum-based input border API with Flutter-aligned `InputBorder`, `OutlineInputBorder`, and `UnderlineInputBorder` classes, including custom sides, radii, gap padding, and borderless input.
- Support state-specific input borders for enabled, focused, error, and disabled states across Material form controls.
- Add independent `DropdownM2.menu_border_radius` configuration and expose the new border classes through the Python SDK and Dart protocol.

Bug Fixes:
- Honor underline corner radii and correctly apply borderless and outlined borders on Cupertino text fields.
- Resolve default border styling from the active Material theme instead of always using black borders.
- Expose fixed Flutter defaults in Python control properties so property reads reflect effective defaults.

Enhancements:
- Keep legacy enum members and loose border properties working through deprecation shims until their planned removal in 1.3.0.
- Preserve nested dataclass fields during message encoding to support nested differ updates.

Documentation:
- Add the InputBorder migration guide, class reference pages, release-note entries, and examples covering border styles, custom styling, per-state borders, and DropdownM2 menu shaping.

Chores:
- Update existing examples to use the class-based border API.


sdk/python/packages/flet/src/flet/controls/material/form_field_control.py
```diff
@@ -15,30 +17,156 @@
...
+
+    Must be non-negative.
+    """
+
+    def __post_init__(self):
+        self._type = "outline"
+
+
+@value
+class _NoInputBorder(InputBorder):
```
FeodorFitsner: I'd make it public: `NoInputBorder`, so you can do `textfield.border = ft.NoInputBorder()`?

ndonkoHenri: Ok., will update.

FeodorFitsner (approved):

PR #6799, opened 2026-08-29
feat(`flet build linux`): app-icon support for Linux bundles

### LINKED ISSUE #2269: `flet build`: add icon for Linux bundle
Discussion and solutions: https://github.com/flutter/flutter/issues/53229#issuecomment-1703233957

ndonkoHenri:
## Summary

Linux was the only platform `flet build` shipped without an app icon: `flutter_launcher_icons` has no Linux generator, so `assets/icon.png` was silently ignored and every built app ran with a generic icon. This adds one, through both mechanisms Linux actually uses.

Closes #2269

Screenshot 2026-08-28 at 8 06 19 PM
<br>
Screenshot 2026-08-28 at 8 06 31 PM

## What changed

**The window icon (X11 / XWayland).** `customize_icons()` resolves `icon_linux.png → icon.png → the template default`, stages it, and CMake installs it into the bundle as `data/app_icon.png`. The runner loads it at startup and sets it as the default window icon, which taskbars and window switchers read via `_NET_WM_ICON`. No installation step required.

**The desktop entry (Wayland).** Wayland has no window-icon protocol GTK 3 can reach — `xdg-toplevel-icon-v1` is GTK ≥ 4.19 client-side and Mutter doesn't implement it — so desktops resolve the icon from an *installed* `.desktop` entry matching the app id. The bundle now ships a ready-to-install freedesktop tree:

```
share/applications/<bundle_id>.desktop
share/icons/hicolor/<size>/apps/<bundle_id>.png
```

The runner also calls `g_set_prgname(APPLICATION_ID)`, matching what upstream Flutter's own runner template has done since flutter#154522, so the Wayland app id and X11 `WM_CLASS` both equal the bundle ID and the window matches its entry.

## Implementation detail

The obvious implementation is a silent no-op for most apps, and it took reading GDK to find out.

`gtk_window_set_default_icon_from_file()` returns **success** while GDK discards the icon. `gdk_x11_window_set_icon_list` caps `_NET_WM_ICON` at `GDK_SELECTION_MAX_SIZE` (262144 CARDINALs) and, when the first icon doesn't fit, breaks out of the loop and calls `XDeleteProperty` instead:

```c
/* silently ignore overlarge icons */
if (size + 2 + width * height > GDK_SELECTION_MAX_SIZE(display))
  break;
...
if (size > 0)  XChangeProperty(... "_NET_WM_ICON" ...);
else           XDeleteProperty(... "_NET_WM_ICON" ...);
```

Anything from 512×512 up is dropped — and the build template's own default icon is 1024×1024, so the naive version would have shipped no icon at all in the common case. The runner now scales to 256×256 before setting it. Largest single icon that fits is 511×511.

## Also fixed along the way

**`--description` never reached the built app.** The value was passed to the template under a key no file consumed (`description`) while every template reads `project_description`, so it always rendered empty — meaning **web builds have shipped a blank `<meta name="description">` and a blank PWA manifest description since the option was introduced**. Both now receive it, as does the new desktop entry's `Comment=`.

Fixing that exposed an injection surface: the value is interpolated into YAML, JSON, HTML and a desktop entry, each with different escaping rules. A description containing an apostrophe produced an unparsable `pubspec.yaml` and broke the build **on every platform**. Each sink is now escaped for its own format, with tests covering quotes, backslashes and newlines.

**`--linux-categories`** sets the desktop entry's `Categories` (default `Utility`), resolving CLI → `[tool.flet.linux].categories` → default.

**CI gap:** `flet-build-test.yml` didn't list `sdk/python/templates/build/**` in its path filters, so build-template changes never triggered the Linux build test — which is how a broken Linux runner could have merged unnoticed.

## Verification/Testing

CI ran a temporary harness (removed before this PR) building a fixture app across four legs — no icon, 256×256, 512×512, and one with quotes/newlines/tabs/backslashes in the description plus a space in the artifact name — on **x86_64 and arm64**. All green, including runtime assertions under Xvfb reading back `WM_CLASS` and `_NET_WM_ICON = 256x256` on every leg.

Then confirmed by hand on an **Ubuntu 22.04 arm64 VM**:

- X11/XWayland: taskbar icon appears with no installation.
- Native Wayland: generic icon until the shipped `share/` tree is installed, correct icon and app name afterwards.
- `WM_CLASS` reports the bundle ID.
- The bundle also packages cleanly into a working **AppImage**.

200 unit tests pass; 570 lines of new tests across `test_build_linux.py` and `test_build_template.py`.

## Docs

New **App icon**, **Application categories** and **Distributing** sections on the Linux page, the latter with annotated AppImage / `.deb` / `.rpm` recipes. `--description` gains a settings section it never had. The apt prerequisite list is now generated from `flet_cli.utils.linux_deps` via a crocodocs partial, so the docs can no longer drift from what CI installs.

## Notes
- Building on arm64 Linux needs a master-channel Flutter and an explicit `flutter precache --linux`.

## Summary by Sourcery

Add complete Linux app icon and desktop integration support while fixing build-description propagation and documenting Linux distribution workflows.

New Features:
- Add Linux app icon support with X11 window icons and installable freedesktop desktop-entry integration for Wayland and application menus.
- Add configurable Linux application categories through the CLI and pyproject.toml.

Bug Fixes:
- Fix build descriptions so they reach generated applications across Linux and web targets and remain valid when containing special characters.
- Ensure oversized Linux icons remain usable by scaling them before applying the window icon.

Enhancements:
- Generate Linux icon theme metadata and desktop integration files as part of the build bundle.
- Generate Linux prerequisite documentation from the canonical dependency list.

CI:
- Trigger the Linux build workflow when build-template files change.

Documentation:
- Document Linux app icons, application categories, distribution recipes, and build descriptions.

Tests:
- Add comprehensive tests for Linux icon staging, desktop-entry rendering and escaping, category resolution, and build-template validity.

Chores:
- Add generated JavaScriptMode WebView documentation and include it in the documentation sidebar.


ndonkoHenri:
Test artifacts (`.AppImage` and `.deb`): https://github.com/flet-dev/flet/actions/runs/33232340069

image

FeodorFitsner (approved):

PR #6803, opened 2026-08-30
Client actions: perform gesture-gated work without a round trip to Python

### LINKED ISSUE #1579: copy to clipboard fails on Safari
On Safari 16.5.1 copy to clipboard fails.  This fails in older releases of flet  as well although it worked before, so this appears to be a safari specific issue

**Code example to reproduce the issue:**

```python
import flet as ft


def test_click(event):
    event.page.set_clipboard("Hello World")


def main(page: ft.Page):
    t = ft.ElevatedButton(text="Copy to clipboard", on_click=test_click)
    page.add(t)


ft.app(target=main, view=ft.WEB_BROWSER)

```

**Describe the results you received:**

Safari show an error in the console: `copy is not successful NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission`

**Describe the results you expected:**
expect the copy to clipboard to succeed

**Additional information you deem important (e.g. issue happens only occasionally):**

apparently  safari will only allow a copy to clipboard on a user action, not an API call...much like it does not allow opening new tabs. see https://stackoverflow.com/questions/62327358/javascript-clipboard-api-safari-ios-notallowederror-message. This was a iOS only issue but now is done on macOS as well.

**Flet version (`pip show flet`)**:

```
0.7.4 and 0.8.1
```

**Operating system**:

**Additional environment details:**

### LINKED ISSUE #3710: FilePicker not opening dialog on web for iOS
### Duplicate Check

- [X] I have searched the [opened issues](https://github.com/flet-dev/flet/issues) and there are no duplicates

### Describe the bug

FilePicker is not opening the dialog to select a file when it's a web app and accessing it from iOS.
I'm testing this form an app I'm making and it works fine when I access the web app from an Android device or laptop, but not from an iPhone.
I thought it might be something with my code, but then I checked your controls gallery and I have the same issue, dialog is not opening if I try it from an iPhone (https://flet-controls-gallery.fly.dev/utility/filepicker)

I tried this with different browsers on the same iPhone (Safari, Chrome, Firefox and Qwant) and with a different iPhone, but I got the same issue.

I also tried setting the desktop mode on the browser, but same result.

### Code sample

https://flet-controls-gallery.fly.dev/utility/filepicker

### To reproduce

1. Access the website https://flet-controls-gallery.fly.dev/utility/filepicker from an iPhone
2. Try to pick a file

### Expected behavior

Dialog to select a file for uploading should open

### Screenshots / Videos

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

https://github.com/user-attachments/assets/9d58b299-1fb4-472d-a78c-ff9c244c439a

### Operating System

macOS

### Operating system details

iOS 17.5.1

### Flet version

0.23.2

### Regression

No, it isn't

### Suggestions

_No response_

### Logs

```console
[Paste your logs here]
```

</details>

### Additional details

_No response_

FeodorFitsner:
Fixes #3710. Fixes #1579.

## The problem

Browsers only let a page open a file picker, write to the clipboard, show a share sheet or open a new tab **while they are still handling the user's click or key press**. Sending that click to Python and acting on the reply takes longer than the permission lasts.

WebKit enforces this and reports nothing; Chrome and Firefox allow the same calls. So an app worked on Android and desktop and silently did nothing on iPhone and iPad — which made a browser rule look like a Flet bug. `FilePicker` looked especially broken because `save_file()` kept working: it clicks a download link, which isn't gated.

`Button.url` already avoided this by being declarative — [`openWebBrowser(url)`](https://github.com/flet-dev/flet/blob/main/packages/flet/lib/src/controls/button.dart) runs inside `onPressed`, before anything reaches Python. This generalizes that.

## The approach

A **client action** names a service method the client invokes itself, synchronously, from the gesture callback:

```python
picker = ft.FilePicker(on_result=handle_result)
page.services.append(picker)

ft.Button("Upload", action=ft.PickFiles(picker, allow_multiple=True))
ft.Button("Copy",   action=ft.CopyToClipboard(token))
ft.Button("Open",   action=ft.OpenUrl("https://flet.dev", target=ft.UrlTarget.BLANK))
ft.Button("Share",  action=ft.ShareText("Check out Flet"))
```

Two findings kept this small. `Control.invokeMethod` calls its listener synchronously when one is registered, and every service's `_invokeMethod` reaches its gated call before its own first `await`. So one dispatcher reuses **all** existing service code — no per-action Dart handler, no registry, no duplicated logic. `FilePickerService._files` and `uploadFiles` are untouched, so `upload()` works exactly as before.

## What's included

**API** — `ft.OpenUrl`, `ft.CopyToClipboard`, `ft.PickFiles`, `ft.ShareText`, and an `action` property on the ten controls that carry `url` today. `ClientAction` is the public base class for typing; the underlying `(service, method, args)` triple stays internal.

`ActionControl` (a `@control(kw_only=True)` base, following `AdaptiveControl`) contributes the property to all ten controls — ~110 lines lighter than repeating it, with one place to edit the docs. `kw_only` matters: without it the inherited field would claim the first positional slot and `ft.Button("Open")` would stop binding to `content`.

**`FilePicker.on_result`** — a gesture-initiated pick has no caller to return to, so the selection arrives as an event. Picked files stay on the `FilePicker`, so they pass straight to `upload()`.

**Fail-fast** — `pick_files()` called from Python on a browser that has already reported it won't open a dialog now raises immediately instead of hanging for its full one-hour timeout. Gated to Apple/WebKit with no live `navigator.userActivation`, because Chrome and Firefox permit programmatic file-input clicks outside a gesture and treating them as blocked would break apps that work today.

**Bug fix: `UrlTarget.BLANK`** — its value was `"blank"` while `SELF`/`PARENT`/`TOP` all carry the spec's leading underscore, so it reached `window.open()` as an ordinary window *name*: the first such link opened a tab called `blank` and every later one reused it. The `LaunchMode.externalApplication` upgrade never fired either. Introduced in #5382, never released.

**Two internal renames** — the protocol enum `ClientAction` → `MessageAction` and `ClientMessage` → `Message`, matching the Dart names they were already documented as mirroring, and freeing `ClientAction` for the public class. Neither was ever exported.

**Docs** — a Client actions cookbook article, reference pages for all new types, and one runnable example per action beside its service's existing examples. The `FilePicker`, `Clipboard`, `Share` and `UrlLauncher` pages each gain a warning explaining why the imperative call does nothing in a web app on iOS.

## Testing

New `tests/test_client_actions.py` (8 tests) covers serialization through `page.add()` rather than the action object alone — that path caught a real bug where the service cache lived in `page._internals`, which **is** sent to the client, so a page holding any action failed to render with `AttributeError: type object 'Clipboard' has no attribute '_i'`. Confirmed the test fails without the fix. Full suite: 255 passed, 8 skipped. Dart analyzer clean.

## Reviewer notes

- **`url` semantics are unchanged.** Rather than lowering `url` to an `OpenUrl` action at serialization, both go through one Dart helper (`runControlActions`) — `openWebBrowser` still leaves all ten controls, but `url` is provably identical because it's the same call.
- **Action arguments are static**, known before the tap. That's the browser's rule, not Flet's; the clipboard example shows updating the action as a value changes.
- **`Clipboard.get` still prompts on Safari** whatever we do. An action makes the read possible, not invisible.
- `url` is still duplicated across 11 controls and could fold into the same base class in a follow-up; the mechanics are now proven.

## Summary by Sourcery

Enable gesture-gated browser operations to run on the client before events reach Python.

New Features:
- Add declarative client actions for opening URLs, copying to the clipboard, selecting files, and sharing text during the originating user gesture.
- Allow supported controls to execute one or more client actions through a new public action property and expose reusable action types for service integrations.
- Add FilePicker.on_result support for gesture-initiated file selection and preserve selected files for subsequent uploads.

Bug Fixes:
- Fail fast when web browsers have already blocked an imperative file-picker request instead of waiting for the request timeout.
- Correct UrlTarget.BLANK to use the standard _blank target and open independent new tabs as intended.

Enhancements:
- Dispatch existing control URLs and new actions through a common synchronous client-side execution path while preserving existing URL behavior.
- Rename internal messaging protocol types to MessageAction and Message to avoid conflict with the public ClientAction API.

Documentation:
- Add a client-actions cookbook, API reference pages, runnable examples, and web-platform warnings for affected services.

Tests:
- Add serialization and control coverage for client actions, including service sharing, file-picker targeting, URL targets, and unset actions.


sdk/python/packages/flet/src/flet/controls/client_action.py
```diff
@@ -0,0 +1,325 @@
...
+    """
+
+    def __post_init__(self) -> None:
+        from flet.controls.services.clipboard import Clipboard
+
+        self._bind(_shared_service(Clipboard), "set", {"data": self.data})
+
+
+@dataclass
+class PickFiles(ClientAction):
```
ndonkoHenri: Will it make sense to group these new client action classes with the existing ones they belong to? 
For example PickFiles in file_picker.py?

FeodorFitsner: Good call - done in 11a9f167.

`PickFiles` → `file_picker.py`, `CopyToClipboard` → `clipboard.py`, `OpenUrl` → `url_launcher.py`, `ShareText` → `share.py`; `ClientAction` stays as the base.

It turned out to be more than cosmetic. The shared module needed four deferred imports inside `__post_init__` purely to dodge an import cycle, plus a `TYPE_CHECKING` block for `FilePicker` - all gone now that the direction is one-way (service module → `client_action`). The nicest one: `ShareText` was reaching across a module boundary for `share.py`'s private `_share_args`, which is now just a local call. And `PickFiles` needs no helper at all, since it targets the user's own `FilePicker`.

`_shared_service`/`_action_field` lost their underscores, as the service modules now import them; their docstrings note they're internal-but-shared.

Public API is unchanged - only the lazy-import map's module paths moved. Verified each module imports first in a fresh interpreter (no cycle in either direction), and the suite is 255 passed / 8 skipped.

PR #6826, opened 2026-09-08
fix(diff): reconcile value-equal child controls in prop fields of re-rendered components

davidlawson:
## Description

A control that a component re-renders **unchanged** loses its events.

`DiffBuilder._compare_dataclasses` (the frozen branch every `@ft.component` re-render goes through) compares Prop-managed fields with `if old is new or old == new: continue`. Control `__eq__` ignores `_i`, so a child that is rebuilt on re-render but compares equal to the previous render's — which is every rebuilt child with a stable handler (module-level function, `use_callback`, a cached lambda) or a cached subtree — is skipped entirely:

- no `_migrate_state`, so the new instance keeps a fresh `_i` the client has never seen,
- no `_parent` stamp,
- no `_dataclass_added`/`_removed` registration, so the session index keeps the *old* instance.

The rendered tree now holds an orphan while the index points at an instance whose `_parent` chain dies with the old tree. The next event the client dispatches to the original `_i` is silently dropped (`Control with ID 12 not found.` at debug level — the `WeakValueDictionary` entry is gone) or, when anything still references the old instance, raises `RuntimeError: Control must be added to the page first` before the handler runs. `Session.__find_live_control` cannot recover it: the live instance carries a different `_i`.

Only Prop-managed fields are affected — `Container.content`, `DataCell.content`, `DataColumn.label`, `DataTable.columns`, and any single-control or list field without a `default_factory`. Structural list fields (`Column.controls`, `DataRow.cells`, `DataTable.rows`) always go through `_compare_values` and were fine, which is why most apps only see this on `content=`-style children. Per-render lambdas mask it further, since they make every rebuilt control unequal.

**Fix:** skip only identical instances (`old is new`). A value that is, or holds, a control is reconciled even when it compares equal — reconciling an equal subtree emits no patch operations, it just migrates `_i`, stamps `_parent` and registers the pair so the session index is refreshed. The `old == new` short-cut is kept for plain values, where it is all that it ever saved.

Found while identity-caching a large `DataTable2` in a component: every parent-driven re-render left the whole table orphaned and the hover handlers on ~1,800 cells raised. Two of the four new tests in `test_object_diff_frozen.py` fail on `main` (the other two guard that identical and genuinely changed children still behave) and all pass with the fix; the full `packages/flet/tests` suite passes (282 passed, 8 skipped).

## Test code

Run, click **Re-render**, then **Click me** — nothing is printed. Without the re-render, `clicked` is printed on every click.

```python
import flet as ft


def on_click(e):
    print("clicked")


@ft.component
def App():
    count, set_count = ft.use_state(0)
    return ft.Column(
        controls=[
            ft.Text(f"re-renders: {count}"),
            ft.Button("Re-render", on_click=lambda e: set_count(count + 1)),
            # a Prop-managed child (Container.content) with a stable handler:
            # every re-render rebuilds it value-equal to the previous one
            ft.Container(content=ft.TextButton("Click me", on_click=on_click)),
        ]
    )


def main(page: ft.Page):
    page.render(App)


ft.run(main)
```

Headless equivalent (what the added tests pin):

```python
from flet.controls.object_patch import ObjectPatch
import flet as ft

def handler(e): ...

old_child, new_child = ft.TextButton("x", on_click=handler), ft.TextButton("x", on_click=handler)
old, new = ft.Container(content=old_child), ft.Container(content=new_child)
old._frozen = True
patch, added, removed = ObjectPatch.from_diff(old, new, control_cls=ft.BaseControl)

assert new_child._i == old_child._i   # main: AssertionError — fresh _i
assert new_child.parent is new        # main: parent is None
assert new_child in added             # main: not registered, index keeps old_child
```

## Type of change

## Checklist

- [x] I signed the CLA.
- [x] New and existing tests pass locally with my changes.
- [x] I have added changelog entries for user-facing changes, if applicable.

## Additional details

Cost: for a value-equal control-holding prop the deep `__eq__` walk is replaced by a reconcile walk of the same subtree (which skips identical children by identity in the keyed list path), so it is roughly a wash; unequal children were reconciled before too. Plain values keep the equality short-cut.

Reproduced on 0.86.2 and 0.86.5; the affected code is unchanged on `main`.

## Summary by Sourcery

Keep rebuilt, value-equal child controls connected to the live session across component re-renders so their events continue to work.

Bug Fixes:
- Reconcile value-equal child controls in Prop-managed fields during frozen diffs so rebuilt controls retain their identity, parent relationships, session registration, and event handling after component re-renders.

Enhancements:
- Preserve the equality short-circuit for plain values while reconciling control-containing values and keeping identical control instances unchanged.

Tests:
- Add frozen-diff coverage for equal single-control and list Prop fields, identical children, and changed children.


FeodorFitsner:
Nice one, thank you!

This bug could be easily hidden behind lambdas or private handlers.

I can do at least two "workarounds" to avoid this bug.

One:

```py
import flet as ft


@ft.component
def App():
    count, set_count = ft.use_state(0)

    def on_click(e):
        print("clicked")

    return ft.Column(
        controls=[
            ft.Text(f"re-renders: {count}"),
            ft.Button("Re-render", on_click=lambda e: set_count(count + 1)),
            # a Prop-managed child (Container.content) with a stable handler:
            # every re-render rebuilds it value-equal to the previous one
            ft.Container(content=ft.TextButton("Click me", on_click=on_click)),
        ]
    )


def main(page: ft.Page):
    page.render(App)


ft.run(main)
```

Two:

```py
import flet as ft


def on_click(e):
    print("clicked")


@ft.component
def App():
    count, set_count = ft.use_state(0)

    def on_button_click(e):
        on_click(e)

    return ft.Column(
        controls=[
            ft.Text(f"re-renders: {count}"),
            ft.Button("Re-render", on_click=lambda e: set_count(count + 1)),
            # a Prop-managed child (Container.content) with a stable handler:
            # every re-render rebuilds it value-equal to the previous one
            ft.Container(content=ft.TextButton("Click me", on_click=on_button_click)),
        ]
    )


def main(page: ft.Page):
    page.render(App)


ft.run(main)
```

In both cases "Click me" button handler is re-created on each handler.

FeodorFitsner (approved):

PR #6829, opened 2026-09-09
fix(web): honour the jsSend transfer list instead of dropping it

ndonkoHenri:
## Problem

The Dart web transport declares and calls a three-argument `jsSend`, passing the packet's own ArrayBuffer as a `postMessage` transfer list so bulk `DataChannel` frames cross the Worker boundary without a structured-clone copy:

```dart
// packages/flet/lib/src/transport/flet_backend_channel_javascript_web.dart
external void jsSend(String appId, JSUint8Array data, JSArray<JSObject>? transferList);
...
jsSend(address, jsBytes, transferList);
```

The only JavaScript definition takes **two** parameters:

```js
globalThis.jsSend = async function (appId, data) {
    app.worker.postMessage(data);      // transfer list never arrives
}
```

So the third argument was silently discarded and every frame was copied. The optimisation the Dart side documents has never actually run.

## Fix

`jsSend` now accepts the transfer list and forwards it, falling back to an empty list when absent:

```js
globalThis.jsSend = async function (appId, data, transferList) {
    app.worker.postMessage(data, transferList ?? []);
}
```

`client/web/` and the build template each carry their own copy of `python.js`, so both change.

## Why detaching the buffer is safe

Transferring detaches the caller's view, so this is only correct if no sender touches the packet afterwards. Both senders allocate a fresh `Uint8List` per call and never read it back:

- `FletBackend._send` — `packages/flet/lib/src/flet_backend.dart:720`
- `ProtocolMuxedDataChannel.send` — `packages/flet/lib/src/transport/protocol_muxed_data_channel.dart:55`

The rationale moves onto the `jsSend` declaration's doc comment, where a reader meets the contract, rather than sitting inline in `send()`.

## Verification

`python.js` ships verbatim into a built web app, so the fixed file was dropped into an existing `flet build web` output and loaded in a browser — no rebuild needed. Every protocol frame now travels with a real transfer list; the app boots, renders its controls and images, and animates an extension control unchanged.

63/63 `packages/flet` tests pass; analyzer clean on the changed file.

## Summary by Sourcery

Enable zero-copy transfer of web transport packets by forwarding their underlying buffers to the Python worker.

Bug Fixes:
- Preserve and transfer web DataChannel packet buffers across the worker boundary to avoid unnecessary structured-clone copies.

Enhancements:
- Align the JavaScript transport and Dart bindings around ownership transfer of converted packet buffers.

Tests:
- Add Node.js coverage for buffer detachment and payload preservation in both shipped web transport files.


FeodorFitsner:
Pushed a follow-up fix in 8f61584a to make the worker transfer work for both JavaScript and Wasm builds.

The original fix correctly forwards the previously ignored transfer list and avoids the worker-boundary copy in JavaScript builds. However, for Dart-allocated packets in Wasm, `packet.toJS` and `packet.buffer.toJS` each create a separate JavaScript copy. The transfer list therefore contains a different buffer from the one backing the message: that unrelated buffer is detached, while the actual message buffer is still copied across the worker boundary.

The follow-up converts the packet only once in Dart:

```dart
jsSend(address, packet.toJS);
```

Both the shipped client and build template now select the actual message buffer in JavaScript:

```js
app.worker.postMessage(data, [data.buffer]);
```

This removes the separate buffer conversion and the Dart-side transfer-list construction. JS builds transfer the existing packet buffer; Wasm builds retain one Dart-to-JavaScript conversion copy, then transfer that converted buffer without another byte copy across the worker boundary. This is not a claim of end-to-end zero-copy on Wasm. The existing internal senders allocate fresh packets and do not access them after sending, so transferring ownership is safe.

Validation: all four added Node MessageChannel regression tests pass for the client/template and full packets/subviews. Separate probes compiled from the updated Dart interop declaration and send method to both JS and Wasm also passed against both JavaScript files: the actual message buffers detached, and the received bytes were intact. These were focused transport checks, not full browser/Pyodide app tests.

Dart documents the compiler-dependent conversion behavior here: [Uint8List.toJS](https://api.dart.dev/dart-js_interop/Uint8ListToJSUint8Array/toJS.html) and [ByteBuffer.toJS](https://api.dart.dev/dart-js_interop/ByteBufferToJSArrayBuffer/toJS.html).

FeodorFitsner (approved):

